PR #9, opened 2022-06-12
Compile high-level programs to low-level representation

rachitnigam:
- construct fsms
- compilation working, externs missing
- add extern syntax
- compilation working with externs
- make liveness annotations optional on ports
- compilation finally works
- err when port dont provide guarantee
- remove when blocks
- add pos information to fsm node
- add correctness for compile
- remove hand-compile
- wish we had functors
- Make port into a struct
- parameterize ports using liveness
- Revert "parameterize ports using liveness"
- Revert "Make port into a struct"
- midend nonsense


rachitnigam:
The tests run into https://github.com/cocotb/cocotb/issues/2950 so have to commit the warning message about the open file descriptors

PR #32, opened 2022-09-01
Enable concurrent transactions

rachitnigam:
This PR enables the testbench to fire off transactions in parallel for a pipeline. In short, this makes it so that a pipeline can process multiple inputs at the same time


rachitnigam:
Okay, added a compilation test for a pipeline with II=1 and latency=3. The test has 7 inputs so we should expect it to take 7+3=10 cycles. However, it currently takes 11 cycles. This is because of the register lifetime encoding trick which requires us to provide a dummy event that is never used. However, the compiler does not know this and ends up creating an FSM with 4 states instead of 3 for the pipeline. I'll open an issue to explain in detail what is going on but for now, we'll roll with this implementation.

PR #116, opened 2023-05-05
Fix debug log when connecting to constant value

UnsignedByte:
Behavior is already correct for constant-value connections, but the debug print unwraps a None value.


rachitnigam:
Ah, we should just get rid of that trace call altogether. It was spuriously added

rachitnigam (changes requested):
I would just get rid of the trace statement entirely. It's spuriously added

rachitnigam:
Hey @UnsignedByte, I ended up removing the `trace!` call in another PR so closing this. Regardless, thanks for taking the time to write this up!

PR #127, opened 2023-05-10
Adds a Pass trait to group filament passes.

UnsignedByte:
Currently contains only the transform function, could maybe add more to it later?


rachitnigam:
Seems like a good idea! There used to be a `Transform` trait in the past that was used for passes like [`lower.rs`](https://github.com/cucapra/filament/blob/main/src/passes/lower.rs#L20) but it got removed. No harm brining this back!

PR #130, opened 2023-05-10
Use integer division

rachitnigam:


tests/compile/reduction-tree.fil
```diff
@@ -15,7 +15,11 @@ comp AddRed[#W, #N, #K]<G:1>(
     } else {
         // Required to make recursive calls type check.
         assume pow2(#K-1) == #N/2;
+        // This is a corollary of #K != 0
         assume #N >= 2;
+        // This is not true in general for integer division but
+        // because #N is a power of 2 it is true.
+        assume #N - #N/2 == #N/2;
```
rachitnigam, line 22: @UnsignedByte another assumption that is potentially ripe for an axiom schema

PR #131, opened 2023-05-10
Implement generic serializer

rachitnigam:


rachitnigam:
The failures are probably happening because of Z3 timeouts :-(

PR #139, opened 2023-05-13
Create automatic assumptions for custom functions.

UnsignedByte:
Recognizes constraints like `#l == pow2(#r)` and generates automatic assumptions for them.


src/core/mod.rs
```diff
@@ -8,14 +8,17 @@ mod loc;
 mod port;
 mod signature;
 mod time;
+mod width_rep;
```
rachitnigam, line 11: This module does not exist. Probably showed up due to a merge error.

src/core/implication.rs
```diff
@@ -0,0 +1,88 @@
+use std::fmt::Display;
+
+use itertools::Itertools;
+
+use crate::utils::{Binding, SExp};
+
+use super::{expr::EvalBool, OrderConstraint, Expr};
```
rachitnigam, line 7: No space between imports

src/core/expr.rs
```diff
@@ -55,7 +113,13 @@ impl std::fmt::Display for UnFn {
         }
     }
 }
+
 impl UnFn {
+    /// Returns the generated assumptions for this function given a left and right.
+    pub fn assume(self, left: &Expr, right: &Expr) -> Vec<Assume> {
+        Into::<FnAssume>::into(self).assume(left, right)
```
rachitnigam, line 120: Avoid using the `Into` trait. Instead, you can do something like:
```
FnAssume::from(self)
```

src/core/expr.rs
```diff
@@ -39,6 +41,62 @@ impl From<Op> for utils::SExp {
...
+        Self {
+            assumptions
+        }
+    }
+
+    /// Creates the assumptions necessary for this function
+    /// Assumes `l = f(r)`
+    fn assume (&self, left: &Expr, right: &Expr) -> Vec<Assume> {
+        let bind = Binding::new(
+        vec![(FnAssume::left(), left.clone()), (FnAssume::right(), right.clone())]
```
rachitnigam, line 88: Please run `cargo fmt` to get rid of the weird formatting

src/core/constraint.rs
```diff
@@ -102,7 +103,13 @@ impl OrderConstraint<Expr> {
         }
     }
 
-    pub fn eval(self, binding: &Binding<Expr>) -> FilamentResult<bool> {
+    pub fn exprs(&self) -> Vec<&Expr> {
+        vec![&self.left, &self.right]
+    }
+}
+
+impl EvalBool for OrderConstraint<Expr> {
```
rachitnigam, line 111: The general recommendation from Rust is that if you're converting from type `A` to `B`, you can implement the `impl From<A> for B` which is a more general, known interface. Maybe it's worth implementing this as a 
```
impl From<OrderConstraint<Expr>> for bool { ... }
```

src/core/expr.rs
```diff
@@ -39,6 +41,62 @@ impl From<Op> for utils::SExp {
     }
 }
 
+/// A custom function definition
+pub struct FnAssume
```
rachitnigam, line 45: More documentation is needed here. Does this represent the axioms for a particular function?

src/core/expr.rs
```diff
@@ -39,6 +41,62 @@ impl From<Op> for utils::SExp {
...
+{
+    assumptions: Vec<Implication<Expr>>
+}
+
+impl FnAssume {
+    /// Creates a unique ID to be replaced for function definitions.
+    /// Returns `(l, r)`: [core::Id] terms for the left and right expressions used in `assume`
+    fn replaceable_ids() -> &'static (Id, Id) {
```
rachitnigam, line 53: Does this exist to allow us to have "pattern variables" in our axioms? For example, if we want to say:
```
x + 1 - 1 = x
```
It'll internally be represented as?
```
_FnAssume_left + 1 - 1 = x
```

src/core/expr.rs
```diff
@@ -72,6 +136,106 @@ impl UnFn {
     }
 }
 
+impl From<UnFn> for FnAssume {
+    /// Returns the default [FnAssume] function assumptions for every [UnFn]
+    fn from(func: UnFn) -> FnAssume {
```
rachitnigam, line 141: Very cool! So this is where all the axioms for a function is specified. Some non-actionable thought: It is currently quite version to express some of the constraints. Maybe we can borrow a page from [egg's pattern language][egg-pl] and build something more concise in the future.

[egg-pl]: https://docs.rs/egg/latest/egg/#simple-example

src/core/constraint.rs
rachitnigam: Ah wait, this requires a binding to be resolved. NVM, it cannot be a `From` implementation

src/core/implication.rs
```diff
@@ -0,0 +1,88 @@
...
+
+/// A type representing the expression a => b
+#[derive(Clone)]
+pub struct Implication<T> {
+    guard: Option<OrderConstraint<T>>,
+    cons: OrderConstraint<T>
+}
+
+impl<T> Implication<T> {
+    pub fn new(guard: OrderConstraint<T>, cons: OrderConstraint<T>) -> Self {
```
rachitnigam, line 17: I would rename `new` and `new_opt` to be `fact` and `implication` since a non-guarded assumptions is fact that should always be true

src/core/implication.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub fn exprs(&self) -> Vec<&Expr> {
+        match &self.guard {
+            Some (g) => [g.exprs(), self.cons.exprs()].into_iter().flatten().collect_vec(),
+            None => self.cons.exprs()
+        }
+    }
+}
+
+impl EvalBool for Implication<Expr> {
+    fn resolve_bool(self, bind: &crate::utils::Binding<super::Expr>) -> crate::errors::FilamentResult<bool> {
```
rachitnigam, line 46: Style nit: Import `Binding`, `Expr`, and `FilamentResult` in the context instead of qualifying them. Normally, if something is in the `super` module, fully importing it is fine. Otherwise, import it to the last module. For example, for something like `crate::foo::bar::baz`, the import should be `use crate::foo::bar` and the use should be `bar::baz(...)`.

The exceptions to this rule is `FilamentResult` and standard library modules.

src/core/implication.rs
```diff
@@ -0,0 +1,88 @@
...
+
+
+impl<T> From<Implication<T>> for SExp
+where
+    SExp: From<OrderConstraint<T>>,
+{
+    fn from(c: Implication<T>) -> Self {
+        match c.guard {
+            // If g then c.cons else true
+            Some(g) => SExp(format!("(ite {} {} true)", SExp::from(g), SExp::from(c.cons))),
```
rachitnigam, line 64: SMTLIB allows you to [specify implication using `=>`](https://microsoft.github.io/z3guide/docs/logic/propositional-logic).

src/passes/assume.rs
```diff
@@ -0,0 +1,63 @@
...
+/// Add default assumptions to the Filament program
+pub struct Assume;
+
+impl Assume {
+    /// Iterate through the parameter constraints in a signature to see if we can add any default function assumptions
+    fn sig(sig: &core::Signature) -> Vec<core::Command> {
+        sig.param_constraints
+            .iter()
+            .filter_map(|c| match c.inner() {
+                OrderConstraint {
```
rachitnigam, line 16: Maybe the `FnAssume` struct should define a `matches` method or something to abstract this away? Might be useful in general.

src/passes/assume.rs
```diff
@@ -0,0 +1,63 @@
...
+            .map(core::Command::from)
+            .collect_vec()
+    }
+
+    fn component(comp: core::Component) -> core::Component {
+        let mut pre_cmds = Assume::sig(&comp.sig);
+        pre_cmds.extend(comp.body.into_iter());
+
+        core::Component {
+            body: pre_cmds,
```
rachitnigam, line 45: One possible problem here is that the `assume` statements are only visible in the body and not the signature which means that if the signature needs a property to be true for it to be valid, this will not work. There is no good solution to this yet because signature bindings don't allow you to `assume` things but something to keep in mind for future extension.

tests/compile/reduction-tree.fil
```diff
@@ -10,17 +10,10 @@ comp AddRed[#W, #N, #K]<G:1>(
...
     if #K == 0 {
         out = in{0};
     } else {
-        // Required to make recursive calls type check.
-        assume pow2(#K-1) == #N/2;
-        // This is a corollary of #K != 0
-        assume #N >= 2;
-        // This is not true in general for integer division but
-        // because #N is a power of 2 it is true.
-        assume #N - #N/2 == #N/2;
```
rachitnigam, line 22: Awesome! Maybe we should also add some tests under `tests/check` to see if assumptions are being instantiated correctly?

rachitnigam (commented):
Thanks for getting this started! Really exciting stuff. Couple of high-level notes:
1. The big thing in this PR is a mechanism to state axioms for functions using a pattern language and "applying them" when we see other constraints in the context. This is a pretty powerful tool, but as you noticed, quite fiddly to make easy to read. In the long-term, we should abstract the process of stating and matching and maybe take ideas from the [egg pattern language][egg-pl].
2. A bunch of the comments are related to Rust style. I'll probably not provide many more of them in the future but useful to get used to them. `cargo fmt` is going to be your friend here!
3. I'm a big fan of doc comments and testing. As a rule of thumb, when adding a new `struct`, `trait`, or `fn`, please add a doc comment for them.

Apart from that, everything is looking pretty good! Since this is your first major PR, couple of onboaring notes for dealing with PRs and the merge process:
1. Usually, once you get comments, try to go through them and mark the ones that have been addressed as "resolved" using Github's "resolve conversation" button
2. If there was a substantial design question that required a rewrite, request a re-review. Since this is your first PR, request a re-review once you're done anyways. In the future, I'll leave some PRs up to your discretion after providing feedback.

Hopefully this all makes sense! Super excited to see this feature land soon!

[egg-pl]: https://docs.rs/egg/latest/egg/#simple-example

rachitnigam:
Oh, also pro-tip: If you're addressing an issue, you can link the PR to the link either by clicking on the "development" pane on the right side of this PR page or using [Github's closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

src/core/expr.rs
UnsignedByte: Yes, you would just use left on both sides and then bind left to x

src/core/expr.rs
UnsignedByte: Yep! This is what I was thinking as well, adding a `From<String> for Expr` would make it much easier to do!

src/passes/assume.rs
UnsignedByte: Created an issue for this at #141 to save it.

rachitnigam:
Also, merge with the `main` branch! I've added a new check that makes sure that the code is correctly formatted

src/passes/assume.rs
rachitnigam: Perfect!

src/core/expr.rs
rachitnigam: Great! We can make an issue to maybe revisit this as our axiom set becomes larger.

Issue #126, opened 2023-05-09
Reimagining Compilation through multiple IRs


rachitnigam:
The current compiler has [exactly one AST][core-ds] that's used throughout the compilation process. There are numerous issues to this approach:
1. The same AST represents high-level operations like `invoke` and low-level operation like `fsm` that show up during compilation
2. Binding information is always annoying to get, especially for passes that mutate the program. We currently use the [binding data structures][binding-ds] to make it easier to resolve some of this information but it only works for analyses and not mutating passes which build their own, ad-hoc binding information (see [bundle-elim](https://github.com/cucapra/filament/blob/main/src/passes/bundle_elim.rs#L14)).
3. No pass can make strict assumptions about the structure of the program. For example, even though we know bindings have been resolved before interval checking runs, we still need to use methods that return `Option` and unwrap them explicitly
4. A corollary of (3) is that sometimes when adding new features, we forget to validate the structure of the new construct and then get weird errors in latter passes.

Thankfully, this is a well-known problem in compilers and there is a known good solution. We should take a hint from [Rust's IRs][rust-irs] and separate out the phases of compilation into different parts that operate on different IRs:
1. A frontend AST representation that allows syntax niceties and doesn't guarantee that the program is well-formed
2. An IR that resolves the binding of all names and guarantees that bindings are well-formed: if there is an identifier in the program, it points to the right kind of data structure for it (invocations, instances, events, bundles etc). 
3. A low-level IR that represents things like FSMs and is very close to Calyx IR. Maybe this should be Calyx IR.

## Frontend AST

I think the current [core data structures][core-ds] are a good candidate for (1). However, there are a couple of things that can be improves. For example, we definitely want to support namespacing in the future (#118). The first PR to implement these changes doesn't have to do it but we should be future proofing things. For reference, this IR should roughly correspond to [Rust's HIR][rust-hir] and roughly follow its ideas on equality and identity of IR nodes.

 A big unknown question whether we should perform type checking at this point and pass that information to the MIR or do that checking in the MIR. The answer to this depends on whether we expect downstream analyses to consume typing information or not. If the overhead is low for doing this, we might as well do it here. In the future, we might want to support type/parameter inference of some form. We can either do it at this level, or define another IR like [Rust's Typed HIR][rust-thir] to support type elaboration.

## Mid-Level IR

[...]

PR #149, opened 2023-05-22
New IR

rachitnigam:


rachitnigam:
This implements an initial version of the IR described in #126. The two missing features are:
1. Constant ports
2. Using interface signals in the body. These are not used super often so not sure if its worth supporting them.

Furthermore, the IR generation panics every time there is a binding problem. We need to figure out if a separate pass figure out binding structures and report errors or should this be the job of `astconv`.

Furthermore, there is no way of deleting things like instances or invocations from a component. Not sure if we want this capability but it can be done using a separate map that tracks validity of things in `IndexStore` objects.

Next steps:
- Rewriting the compiler phases (type checking, interval checking, lowering, etc.) using the new IR.
- Adding enough debug information to the `ir::Component` object so that we can report good error messages. We should store things like location information using a bunch of `DenseInfoMap`s that are stored behind a pointer so that we don't increase the size of `ir::Component` too much more.
- Figure out how lower will work

Inevitably, as we work on these, the IR's design will change: we'll add more methods on all the IR objects and maybe add more information to things. This is fine an expected!

rachitnigam:
@UnsignedByte we can chat more about it but if you're curious about what the IR looks like, you can run `cargo run -- <file> --show-ir` and it'll generate a printed representation of the IR for the program.

sampsyo:
Awesome! Predictably, I think this is an interesting implementation problem:

> Furthermore, there is no way of deleting things like instances or invocations from a component. Not sure if we want this capability but it can be done using a separate map that tracks validity of things in `IndexStore` objects.

Hopefully it's possible to avoid needing deletion as long as possible, but when it is necessary, this (a "soft delete" map) doesn't sound like a bad way to go. Another (possibly very bad) option is to make such updates functional: if you want to delete stuff, you have to do the moral equivalent of a functional-programming `filter` and generate a new component that drops some of the contents.

PR #151, opened 2023-05-24
bump calyx version

rachitnigam:


rachitnigam:
@UnsignedByte and @gabizon103 this will break things for you since we renamed the Calyx `fud` state from `futil` to `calyx`. Fix it requires updating the Calyx repo and reinstalling `fud`. Once that is done `fud check` should report that everything is correct and you'll be good to go.

PR #162, opened 2023-06-01
Print assignments demonstrating false `assert`s

rachitnigam:
Uses the solver to get an assignment that demonstrates that the assert is false.


rachitnigam:
Additionally add information tracked by components that can be used to report user-facing error messages.

PR #168, opened 2023-06-07
Explicitly print expressions in IR printer

gabizon103:
Exprs were getting printed in the IR as `%e0` (and so on), so now we print what they actually are. Some examples:

1. In a component signature, we were previously printing `comp %comp1[...]<%ev0: %e0>(...)`, but now it looks like `comp %comp1[...]<%ev0: 1>(...)` (or whatever the delay of the event is). The same change happens in component bodies, where we print out events.
2. In instantiations, we were previously printing `%inst0 = instance %comp0[%e0];`, but now it looks like `%inst0 = instance %comp0[#W];`.

### DIFF
--- a/src/ir/printer.rs
+++ b/src/ir/printer.rs
@@ -241,18 +241,10 @@ impl Printer<'_> {
         {
             match pos {
                 Position::First((idx, ev)) | Position::Middle((idx, ev)) => {
-                    write!(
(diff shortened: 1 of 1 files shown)


rachitnigam:
Awesome! LGTM (looks good to me)! Merge it

PR #171, opened 2023-06-09
Small frisc implementation

gabizon103:
Wrote a RISC-V 32I instruction decoder, ALU, and register file. 
These are all connected in the `cpu` module, which takes an instruction as an input and outputs the ALU inputs and result for each instruction (this made it pretty easy to verify everything was working correctly). 
I added some tests for each component to the `runt` config as well. 
Next steps are to add support for branching and interfacing with memories.


frisc/src/extras/extras.fil
```diff
@@ -0,0 +1,40 @@
...
+  ) where L > G, #WIDTH > 0;
+
+  comp Ternary[#WIDTH]<G:L-(G),?L:1=G+1>(
+    @[G,L] guard:1,
+    @[G,L] opTrue: #WIDTH,
+    @[G,L] opFalse: #WIDTH
+  ) -> (
+    @[G,L] out: #WIDTH
+  ) where L > G, #WIDTH > 0;
+}
```
rachitnigam, line 40: The last three seems like they should be implementable within Filament itself. Any reason to implement them in Verilog? Also, might be worth making the `onehotmux` a parameterized primitive one day (not necessarily in this PR)

frisc/src/cpu.fil
```diff
@@ -0,0 +1,30 @@
+import "frisc/src/alu.fil";
+import "frisc/src/decode.fil";
+import "frisc/src/regfile.fil";
+import "frisc/src/extras/extras.fil";
+
+comp cpu<G:1> (
```
rachitnigam, line 6: One random, completely undocumented, naming scheme I abide by is CamelCasing component names so that they look distinct from ports and invocations. Also, somehow, they feel similar to classes since we have to instantiate them using `new`.

frisc/src/regfile.sv
```diff
@@ -0,0 +1,26 @@
...
+  input wire logic en,
+  input wire logic [4:0] rin1,
+  input wire logic [4:0] rin2,
+  input wire logic [4:0] rd,
+  input wire logic [31:0] data,
+  output reg [31:0] rs1,
+  output reg [31:0] rs2
+);
+  integer i;
+  reg [31:0] rf [31:0];
```
rachitnigam, line 13: Prefer using `logic` instead of `reg`

frisc/src/extras/extras.fil
gabizon103: For the shifter, I wanted to use Verilog's right arithmetic shift operator (`>>>`). For the other two, I didn't have a specific reason I implemented them in Verilog, so I'll rewrite them in Filament.

frisc/src/cpu.fil
gabizon103: I like this, and also realized I was not deliberate with my naming scheme here at all so I'll change it to this

frisc/src/regfile.sv
gabizon103: Will do, thanks!

frisc/src/alu.fil
```diff
@@ -0,0 +1,51 @@
+import "primitives/core.fil";
+import "frisc/src/extras/extras.fil";
+
+// ALU
+
```
rachitnigam, line 5: No space needed here

frisc/src/alu.fil
```diff
@@ -0,0 +1,51 @@
...
+  aluMinus32 := new Select[33,32]<G>(aluMinus.out);
+  
+  ltguard := new Xor[1]<G>(inA31.out, inB31.out);
+  ltTernary := new Mux[1]<G>(ltguard.out, inA31.out, aluMinus32.out);
+  lt := new ZeroExtend[1,32]<G>(ltTernary.out);
+  ltu := new ZeroExtend[1,32]<G>(aluMinus32.out);
+
+  mux := new OneHotMux[32]<G>(funct, selAddSub.out, sll.out, lt.out, ltu.out, xor.out, sr.out, or.out, and.out);
+  out = mux.out;
```
rachitnigam, line 50: Awesome! I think this can serve as a good example for why we need more frontend magic to make our code look more readable. Great job pushing through it anyways!

frisc/src/decode.fil
```diff
@@ -0,0 +1,148 @@
...
+  @[G, G+1] isAUIPC:1,
+  @[G, G+1] Uimm:32,
+  @[G, G+1] Iimm:32,
+  @[G, G+1] Simm:32,
+  @[G, G+1] Bimm:32,
+  @[G, G+1] Jimm:32,
+  @[G, G+1] funct:8,
+  @[G, G+1] isAdd:1,   // 1 if add, 0 if sub
+  @[G, G+1] isArith:1, // 1 if arith, 0 if log
+  @[G, G+1] rf_write:1
```
rachitnigam, line 26: Really jealous of Spade's `enum` and `match` commands right now

frisc/src/extras/extras.fil
rachitnigam: Good point! We should have an arithmetic shift operation in the standard library as well!

frisc/src/extras/extras.sv
```diff
@@ -0,0 +1,50 @@
...
+module RightShifter #(
+  parameter WIDTH = 32
+) (
+  input wire logic signed [WIDTH-1:0] in,
+  input wire logic [4:0] shamt,
+  input wire logic isArith, // 0 for logical, 1 for arithmetic
+  output logic [WIDTH-1:0] out
+);
+  assign out = isArith ? in >>> shamt : in >> shamt;
+endmodule
```
rachitnigam, line 39: Another approach to moving this into Filament: add arithmetic shift to the standard library and then implement this component in Filament using `ShiftRight` and `ArithShiftRight`

frisc/src/extras/extras.sv
```diff
@@ -0,0 +1,50 @@
...
+
+module Ternary #(
```
rachitnigam, line 41: Looks like this module is no longer used?

frisc/test/cpu.txt
```diff
@@ -0,0 +1,23 @@
+JSON files don't like when you put comments there, so the test cases are explained here. 
+The start of a new test case is indicated by the reset signal going high.
+The way tests are verified now is that the output of the ALU is wired to the output of the CPU, so we can inspect that.
```
rachitnigam, line 3: Argh, that's annoying. Perhaps we can change our JSON parsing logic to support comments using [another JSON library](https://stackoverflow.com/questions/29959191/how-to-parse-json-file-with-c-style-comments)

rachitnigam (approved):
Looking good! I've left a couple of final comments and moved this to "merge at your discretion" status. For some low-level comments, like using more Filament components, try to address them before merging. For other, more involved issues (like supporting comments in JSON), I recommend opening an issue to track them and addressing them in a future PR.

Great job on this! Super excited to see where it goes next

PR #174, opened 2023-06-13
FFT generator

UnsignedByte:
Builds an FFT generator, includes floating point library.


rachitnigam:
So, this PR is trying to add three different things:
1. Default event parameters
2. Floating point library
3. FFT generator (which seems empty currently)

Can you separate the code out into different PRs so its easier to review this? Normally, we like to keep PRs small so they address one feature and are easy to review.

UnsignedByte:
Force pushed here because it was too far from `main` to easily merge.

UnsignedByte:
@rachitnigam floating point tests fail currently due to floating point error from slightly different arithmetic operations used in both implementations. Should we change `gen_float`'s `check` to allow for an approximation mode (difference under epsilon)?

rachitnigam:
Hm, I thought there was also a bug causing the problem (https://github.com/cucapra/filament/issues/199)? Is this a different problem?

UnsignedByte:
Sorry, that's my bad. I meant that the FFT tests were failing, not that the floating point ones were. It makes sense that the FFT ones are slightly off because the algorithm for the two FFTs is probably slightly different, and thus we should be comparing the floats to a certain epsilon, not directly as raw values.

rachitnigam:
I think that would require a custom tester. Maybe we can define that in [the harness](https://github.com/cucapra/filament/blob/main/fud/harness/utils.py) and expose that somehow? Alternatively we can build a simple tool that ingests two JSON files, interprets the values as floats, and compares them with an epsilon.

UnsignedByte:
I like the second case; what if we change the current `gen_float.py`'s checking script to treat values as floats (under a flag)? We can set `EPSILON = 0` as the default to ensure that current tests still work as normal, but in the case of the FFT we can set epsilon to a normal epsilon value. This would also require the tester to support some packed bus functionality (as I would rather not have `16` outputs per FFT type) but that should be relatively simple, I'm thinking splitting `width` simply into two params, `w` the per number and `n`, number of numbers in the bus, so all current tests can simply look like `32 1` instead of `32`.

rachitnigam:
To be clear, we don't need to have the old test files working exactly the same. If we need to pass another parameter, it's fine to update them

rachitnigam:
@UnsignedByte do we have path towards getting this merged? What remains to be done in the simplest, non-pipelined module? Is it the testing stuff?

UnsignedByte:
> @UnsignedByte do we have path towards getting this merged? What remains to be done in the simplest, non-pipelined module? Is it the testing stuff?

Yeah, just got the testing stuff in but there's a weird bug that I think might either be a verilator issue or some even weirder problem that I'm currently still looking into.

UnsignedByte:
NVM, it looks to be a floating point addition issue with `0+0`.

primitives/signed.fil
```diff
@@ -1,12 +1,30 @@
...
 
   comp SAdd[WIDTH]<'G: 'L-('G), ?'L: 1='G+1>(
      left: ['G, 'L] WIDTH,
      right: ['G, 'L] WIDTH
   ) -> (
      out: ['G, 'L] WIDTH,
   ) where 'L > 'G;
+}
+
+comp SignedConst[WIDTH, VALUE]<'G: 1>() -> (
```
rachitnigam, line 18: What's the difference between signed const and neg const?

tests/floating-point/fft/fft-comb-8pt.fil
```diff
@@ -0,0 +1,209 @@
...
+  twiddle{13} = twiddle{0};
+
+  TR7 := new Const[32, 1060439283]<'G>();
+  twiddle{14} = TR7.out;
+  twiddle{15} = twiddle{2};
+
+  // Stage 0 input (bit reversal on pairs)
+  bundle s0[16]: ['G, 'G+1] 32;
+  s0{0} = in{0}; // 000 -> 000 (0 -> 0)
+  s0{1} = in{1};
```
rachitnigam, line 49: Any point trying to shorten this using `s0{0..2} = in{0..2}` and so on?

primitives/signed.fil
UnsignedByte: Signed const allows us to take an unsigned `N` bit parameter value and convert it to the signed wire equivalent. `NegConst` just takes `x` and returns `-x` I think.

primitives/fixed.fil
```diff
@@ -0,0 +1,12 @@
...
+ * Defines combinational fixed point math operations
+ */
+
+extern "fixed.sv" {
+  comp FXPMult[W, ?D=W/2]<'G: 'L-('G), ?'L: 1='G+1>(
+    left: ['G, 'L] W,
+    right: ['G, 'L] W,
+  ) -> (
+    out: ['G, 'L] W
+  ) where 'L > 'G, D < W, W > 0;
```
rachitnigam, line 11: Why do we need an extern for this? Fixed point multiplication can be implemented in terms of standard multiplication right. The Calyx primitives library implements `std_mult_pipe` (integer mult) in `std_fp_mult_pipe` (fixed point mult): https://github.com/cucapra/calyx/blob/master/primitives/binary_operators.sv#L366

primitives/float.fil
```diff
@@ -119,14 +119,14 @@ comp FAdd[WIDTH, EXPONENT, MANTISSA]<'G: 1> (
 
     // renormalization
     bundle shift_amts[MANTISSA+1]: for<i> ['G, 'G+1] WIDTH;
-    _mant_1_const := new Const[WIDTH, MANTISSA+2]<'G, 'G+1>(); // constant value set if all mantissa bits are 0
```
rachitnigam, line 122: Might be worth giving local names to terms like `MANTISSA+2`. Also, in a follow up PR, we should rename things to follow the naming convention (#292)

primitives/comb.sv
```diff
@@ -31,6 +31,15 @@ module Sub #(
   assign out = left - right;
 endmodule
 
+module Neg #(
+  parameter WIDTH = 32
+) (
+  input wire logic [WIDTH-1:0] in,
+  output wire logic [WIDTH-1:0] out
```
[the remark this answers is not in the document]
UnsignedByte: Shouldnt + and - be the same regardless of signedness?

primitives/fixed.fil
UnsignedByte: Wouldn't we still have to do the right shift at the end to discard the bottom `d` bits? Not sure how verilog can do fixed point multiplication.

UnsignedByte: It doesn't have to be a primitive I guess, I could implement it as a component instead.

tests/floating-point/fft/fft-comb.fil
```diff
@@ -0,0 +1,178 @@
...
+    sinbits := new SignedConst[W, sin_bits(i+pow2(NStages-1), NPoints)]<'G>(); // equivalent to -sin
+    twiddle{i*2} = cosbits.out;
+    twiddle{i*2+1} = sinbits.out;
+  }
+
+  // bit reverse the input
+  bundle bitrev[NPoints*2]: ['G, 'G+1] W;
+  for j in 0..NPoints {
+    let br = bit_rev(j, NStages);
+    assume br < NPoints;
```
rachitnigam, line 30: We should bring back post monomorphization checking for `assume` and `assert` statements. Stuff like this makes me a little nervous although seems obviously true.

tests/floating-point/fft/fft-cooley-tukey.fil
```diff
@@ -0,0 +1,123 @@
+import "primitives/float.fil";
```
rachitnigam, line 1: Did you mean to commit this file? The components don't seem to be implemented yet. I assuming this is for the follow up PR with the pipelined FFT.

tests/floating-point/fft/harness.fil
```diff
@@ -0,0 +1,109 @@
+import "fft-comb-8pt.fil";
+import "primitives/reshape.fil";
+import "fft-comb.fil";
+
+extern "spiral-fft.v" {
```
rachitnigam, line 5: Oh awesome! We have differential testing for this stuff!!!

rachitnigam (approved):
This is really cool stuff! I've left a bunch of comments that can all be resolved at your discretion before the merge. The differential testing thing is particularly exciting to me because then we can do a head-to-head resource comparison between Spiral-generated designs and our own once we have the pipelined implementation.

primitives/signed.fil
rachitnigam: One seems more expressive than the other in that case. We should remove them accordingly and pick one.

primitives/fixed.fil
rachitnigam: Yeah, take a look at how Calyx does this. Fewer things as primitives is better because then Filament can make strong guarantees about the program.

Issue #178, opened 2023-06-15
Compiling programs without main components


gabizon103:
Compiling a Filament program does not currently emit any Calyx code if it does not have a `main` component. If you want to compile and test it, it requires making a fake `main` component that has the same inputs and outputs, wired up to each other. To change this, we could have the user pass in an argument called `toplevel` (that defaults to `main`) when invoking the compiler that tells it where to enter the program.

rachitnigam:
@gabizon103, if this was fixed by #180, let's close it.

Issue #179, opened 2023-06-15
Automatically transform and run RISC-V program with Calyx harness


gabizon103:
With the harness recently working, the only way to "run" a program with the core is to manually put the instructions into the instruction RAM by editing the data file. We should probably change this so that we can use `fud` to read a file containing a RISC-V program, load the appropriate data into the instruction RAM, and then execute it.

Issue #173, opened 2023-06-12
Harness for interfacing with Calyx memories


gabizon103:
Opening an issue to keep track of this, since it is what I'm currently working on and it seems important for general Filament use as well.

My current idea is to compile a Filament design to Calyx and instantiate it, along with whatever memories it needs to interact with, in a top-level Calyx program. I'm not sure how well this approach would scale/if there are more convenient approaches (like adding standard memory components to Filament that get compiled to Calyx memories?), but I'll post updates/discussions here.

rachitnigam:
Awesome! One deeper, language-level problem here is that every port in Filament represents a "stream" of data and Filament programs transform the streams in some way. Here, we seem to want an outer "controller" that decides how to send the data for the stream. 

Maybe we can do this within the language itself? Spade uses the `entity` abstraction (which I don't deeply understand) to represent such controllers. The other question is that such controllers probably don't have the same kind of type system as Filament in general

gabizon103:
I've been working with this quite a bit and I think there is a strong argument for adding this to the language itself. It feels like you do all this work writing things in Filament so that the timing is correct, but then that sort of goes out the window when you want your design to use any memories. Would it make sense to try writing a Filament signature for the Verilog implementation of the Calyx memories?

rachitnigam:
Absolutely! You may even consider trying to implement a multi ported Verilog memory and giving it a filament. I’d be curious to see how it works and if we can do the whole darn thing in filament!

PR #180, opened 2023-06-15
Calyx harness for frisc

gabizon103:
Implements one approach to #173. Now there is a harness we can use for interfacing between the `frisc` core and memories. I compiled the core to Calyx and then instantiated it in a top-level Calyx module that also contains the memories. Every cycle, an instruction is read from the instruction RAM and passed into `CPU`, and the output of the `CPU` is written to a separate memory for the sake of verifying the results. I think this approach works reasonably well, but I want to keep #173 open because it would still be cool to explore doing this within Filament.
Next steps are to use `fud` to automatically load a RISC-V program into the instruction RAM, outlined in #179.


frisc/harness/frisc.futil
```diff
@@ -0,0 +1,354 @@
+extern "../src/extras/extras.sv" {
```
rachitnigam, line 1: It seems we're committing generated code here which, in general, should be avoided so that the tests can rebuild all the artifacts and make sure things work. I'm assuming this is because of #178 but is there a workaround that we can use in the meantime? Or, if #178 is easy to solve, solving it before merging this?

gabizon103: Sure, I'll try to solve #178 first

gabizon103:
Now closes #178

src/frontend/parser.rs
```diff
@@ -734,6 +734,7 @@ impl FilamentParser {
                     imports: imps,
                     externs: vec![],
                     components: vec![],
+                    toplevel: "".to_string(),
```
rachitnigam, line 737: Should this be main?

rachitnigam (approved):
One question but otherwise merge-worthy!

src/frontend/parser.rs
gabizon103: Yes I think it should, thank you for catching that!

PR #181, opened 2023-06-17
Implement optional parameters

UnsignedByte:
Creates optional parameters feature as well as adding it to a number of elements in the primitive library.


rachitnigam (approved):
Looks good! One suggestion before merging: can we add a test for when the program doesn't provide enough parameters in `tests/error/`?

PR #182, opened 2023-06-18
IR Assumption pass

UnsignedByte:
Implements the assumption pass in the IR version of the compiler.


src/ir_passes/assume.rs
```diff
@@ -0,0 +1,174 @@
...
+        match p {
+            ir::Prop::Cmp(ir::CmpOp {
+                op: ir::Cmp::Eq,
+                lhs,
+                rhs,
+            }) => {
+                let exprs = comp.exprs();
+                let funcmatch = match (exprs.get(*lhs), exprs.get(*rhs)) {
+                    (ir::Expr::Fn { op, args }, _) => {
+                        debug_assert!(args.len() == 1, "Currently Unimplemented: {} requires {} arguments, automatic assumptions only implemented for single argument functions.", op, args.len());
```
rachitnigam, line 99: This should be just an `assert!` since it wrong even in release mode.

src/ir_passes/assume.rs
```diff
@@ -0,0 +1,174 @@
...
+                            Binding::new(vec![
+                                (FnAssume::left(), *lhs),
+                                (FnAssume::right(), args[0]),
+                            ]),
+                        ))
+                    }
+                    _ => None,
+                };
+
+                match funcmatch {
```
rachitnigam, line 121: This should use Rust's `let`-`else` pattern instead:
```
let Some((op, binding)) = funcmatch { return Action::Continue };
...
```

src/ir_passes/assume.rs
```diff
@@ -0,0 +1,174 @@
...
+                        )
+                    }
+                }
+            }
+            ir::Prop::And(lhs, rhs) => {
+                let (lhs, rhs) = (*lhs, *rhs);
+                let laction = Assume::prop(lhs, info, comp);
+                match laction {
+                    Action::AddBefore(cmds) => {
+                        match Assume::prop(rhs, info, comp) {
```
rachitnigam, line 145: I'm not sure what's going on here? Why does this need to call a function that returns a Visitor action?

rachitnigam (changes requested):
This needs more comments in the code to explain what's going on before we can merge it

UnsignedByte:
> This needs more comments in the code to explain what's going on before we can merge it

Yep I was just working on that before I went to dinner 👍

src/ir_passes/assume.rs
UnsignedByte: Yeah I'm probably gonna restructure this looking back, the adds should be moved elsewhere.

rachitnigam:
Okay! Let me know when you want me to review it

src/ir_passes/assume.rs
```diff
@@ -0,0 +1,164 @@
...
+
+/// Generates default assumptions to the Filament program for assumptions using custom functions
+#[derive(Default)]
+pub struct Assume;
+
+impl ir::Component {
+    /// Interns new [ir::Expr]s generated from an [ast::Expr] given a binding on [ir::ExprIdx]
+    fn add_expr(
+        &mut self,
+        expr: ast::Expr,
```
rachitnigam, line 16: We shouldn't be working with `ast` in this directory. Once things are in the `ir`, we should only construct `ir` objects.

src/ir_passes/assume.rs
```diff
@@ -0,0 +1,164 @@
...
+                    }
+                    _ => None,
+                } {
+                    log::debug!("Generating default assumptions for {p}");
+                    // Creates the binding for the left and right [ExprIdx]s
+                    let binding = Binding::new(vec![
+                        (FnAssume::left(), left),
+                        (FnAssume::right(), right),
+                    ]);
+                    FnAssume::from(op.clone())
```
rachitnigam, line 127: I see, the `ast` dependence comes from here because `FnAssume` works on `ast` objects. It would be annoying but I think it would be worth duplicating the code since we're going to delete the passes from the non-IR pass once the transition is complete.

src/ir_passes/assume.rs
```diff
@@ -0,0 +1,133 @@
...
+                    lhs.equal(
+                        rhs.sub(one, self).pow2(self).mul(two, self),
+                        self,
+                    ),
+                    self,
+                ),
+                // #l = 1 => #r = 0
+                lhs.equal(one, self).implies(rhs.equal(zero, self), self),
+                // #r = 0 => #l = 1
+                rhs.equal(zero, self).implies(lhs.equal(one, self), self),
```
rachitnigam, line 40: Ha! Isn't this syntax much nicer. If only we had [Scala's context parameters](https://docs.scala-lang.org/scala3/book/ca-context-parameters.html); we could make the `self` argument hidden too and it'd be perfect.

rachitnigam (approved):
Awesome! Merge it!

PR #184, opened 2023-06-18
Add parametrized floating point library

UnsignedByte:


primitives/float.fil
```diff
@@ -0,0 +1,228 @@
...
+) where
+    #WIDTH == #EXPONENT + #MANTISSA + 1,
+    #EXPONENT > 1,
+    #MANTISSA > 1
+{
+     //// comb stage 1 ////
+    left := new _FDecompose[#WIDTH, #EXPONENT, #MANTISSA]<G, G+1>(left);
+    right := new _FDecompose[#WIDTH, #EXPONENT, #MANTISSA]<G, G+1>(right);
+
+    _one := new Const[1, 1]<G, G+1>();
```
rachitnigam, line 60: Why start these names with an underscore?

primitives/comb.fil
```diff
@@ -143,19 +143,19 @@ extern "comb.sv" {
     @[G, L] out: 1,
   ) where L > G, #WIDTH > 0;
 
-  comp ShiftLeft[#WIDTH]<G: L-(G), ?L: 1=G+1>(
+  comp ShiftLeft[#WIDTH, ?#SHIFT_WIDTH=#WIDTH, ?#OUT_WIDTH=#WIDTH]<G: L-(G), ?L: 1=G+1>(
```
rachitnigam, line 146: Looks like this depends on #181 to be merged so let's merge that first before merging this.

rachitnigam (approved):
Looks good but also looks like this depends on https://github.com/cucapra/filament/pull/181 to be merged so let's merge that first before merging this.

primitives/float.fil
UnsignedByte: Components underscored because of a python convention to signify private names, done with actual invokes and instantiations just to help me figure out what exactly should correspond to the verilog.

PR #186, opened 2023-06-21
IR lower pass

UnsignedByte:
Compiles the ir directly into calyx.

Merges the old ast Lowering pass and Compile pass into one, generating FSMs for events and generating guards for filament assignments.

External interface information (input port names, output port names, parameter names) are saved only for externals and the toplevel component.


reticle/data.json
```diff
@@ -0,0 +1,716 @@
+{
```
rachitnigam, line 1: This file shouldn't be committed

runt.toml
```diff
@@ -30,6 +30,15 @@ cmd = """
...
 """
 
+# [[tests]]
+# name = "ir-compile"
+# paths = [
+#   "tests/compile/*.fil"
+# ]
+# cmd = """
+# fud e -s cocotb.data {}.data --to cocotb-out {} -s calyx.flags ' -d canonicalize' -s filament.flags ' --show-models --ir' -q
+# """
```
rachitnigam, line 40: Why is this commented out?

src/ir/comp.rs
```diff
@@ -44,15 +49,15 @@ pub struct Component {
 
     /// Information tracked by the component
     info: IndexStore<Info>,
-    /// Is this an external component
-    pub is_ext: bool,
+    /// `Some(filename, componentname)` if this is an external component, `None` otherwise.
```
rachitnigam, line 52: This doesn't seem to be a tuple

rachitnigam:
We could separate out the parametric programs into a different directory and make them use the original flow? Once monomorphization works, we can switch them back over

runt.toml
UnsignedByte: Most tests do not pass due to state.fil or parameters

src/ir/comp.rs
```diff
@@ -234,12 +261,24 @@ impl Component {
...
+    /// Unannotated Ports in the component
+    pub fn unannotated_ports(&self) -> Vec<(&ast::Id, u64)> {
+        self.info
+            .iter()
+            .filter_map(|(_, info)| match info {
+                Info::UnannotatedPort { name, width } => Some((name, *width)),
+                _ => None,
+            })
+            .collect()
+    }
```
rachitnigam, line 281: This will be a costly operation since `self.info` can get quite large.

src/ir/from_ast/astconv.rs
```diff
@@ -223,8 +223,18 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
     ) -> ParamIdx {
         let default = param.default.as_ref().map(|e| self.expr(e.clone()));
         let info = self.comp.add(ir::Info::param(param.name(), param.pos()));
+        let is_sig_param = ir::ParamOwner::Sig == owner;
```
rachitnigam, line 226: Prefer `matches!` over `==`

rachitnigam: Actually, this should probably be a method on `param` or something

src/ir/from_ast/astconv.rs
```diff
@@ -268,6 +278,14 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
             has_interface: interface_port.is_some(),
         };
         let idx = self.comp.add(e);
+
+        // If the component is expecting interface information and there is an interface port, add it.
+        if let (Some((name, _)), Some(src)) =
+            (interface_port, &mut self.comp.src_info)
```
rachitnigam, line 284: Why is this a mutable reference?

src/ir/from_ast/astconv.rs
```diff
@@ -280,6 +298,7 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
     }
 
     fn port(&mut self, pd: ast::PortDef, owner: ir::PortOwner) -> PortIdx {
+        let is_sig_port = matches!(owner, ir::PortOwner::Sig { .. });
```
rachitnigam, line 301: Should be a method on owner

src/ir/from_ast/astconv.rs
```diff
@@ -355,8 +374,17 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
         let param = self.comp.get_mut(p.live.idx);
         param.owner = ir::ParamOwner::bundle(idx);
 
+        // If this is a signature port, try adding it to the component's external interface
+        if is_sig_port {
```
rachitnigam, line 378: Inline the definition of the variable and the use since it isn't that complicated an expression

src/ir/from_ast/astconv.rs
```diff
@@ -854,58 +883,72 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
-        .collect_vec();
     // chains external signatures and component signatures (in post-order) to get all signatures associated with this namespace
-    let signatures = ns
-        .externals()
-        .map(|(_, sig)| sig)
-        .chain(components.iter().map(|(_, c)| &c.sig));
+    // let signatures = ns
+    //     .externals()
+    //     .map(|(_, sig)| sig)
+    //     .chain(components.iter().map(|(_, c)| &c.sig));
```
rachitnigam, line 908: Commented code left here

src/ir/from_ast/astconv.rs
```diff
@@ -854,58 +883,72 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
+    let mut dummy = ir::Component::new(false);
+    for cidx in order {
+        let comp = std::mem::take(&mut ns.components[cidx]);
+        log::debug!("Compiling component {}", comp.sig.name);
+        let idx = sig_map.get(&comp.sig.name).unwrap().idx;
+
+        std::mem::swap(ctx.get_mut(idx), &mut dummy);
+        BuildCtx::comp(&ctx, &mut dummy, comp, &sig_map);
+        // swap the component back into place
+        std::mem::swap(ctx.get_mut(idx), &mut dummy);
```
rachitnigam, line 951: This is a particularly gnarly sequence of `std::mem` operations. Can we remove this by carefully passing a `&mut` reference to. the component?

src/ir/utils.rs
```diff
@@ -225,38 +249,156 @@ impl<T, V> std::ops::Index<Idx<T>> for DenseIndexInfo<T, V> {
     }
 }
 
+pub struct Traversal {
```
rachitnigam, line 252: Any `pub` thing needs a comment about what it does

src/ir/utils.rs
```diff
@@ -225,38 +249,156 @@ impl<T, V> std::ops::Index<Idx<T>> for DenseIndexInfo<T, V> {
...
+        }
+
+        for (idx, comp) in comps.iter() {
+            for cmd in &comp.cmds {
+                Traversal::process_cmd(&ctx, *idx, cmd, &mut ts);
+            }
+        }
+
+        let order: Vec<_> = ts.collect();
+        debug_assert!(
```
rachitnigam, line 280: Just use `assert!` instead of `debug_assert!`

src/ir/utils.rs
```diff
@@ -225,38 +249,156 @@ impl<T, V> std::ops::Index<Idx<T>> for DenseIndexInfo<T, V> {
...
         let c_resolved = ctx.get(self.owner);
         Foreign {
-            key: f(self.key, c_resolved),
+            key: f(c_resolved, self.key),
             owner: self.owner,
         }
     }
+
+    /// Runs a function used to unwrap the foreign type into a different type.
+    pub fn apply<F, X>(&self, ctx: &impl Ctx<C>, mut f: F) -> X
```
rachitnigam, line 389: Add a comment saying that you probably shouldn't use this if `X` is an `Idx<T>` type since using that index is going to be unsafe.

src/ir_passes/compile.rs
```diff
@@ -0,0 +1,754 @@
...
+    /// Converts an [ir::ExprIdx] into a [calyx::Width].
+    /// Expects the [ir::ExprIdx] to either be a singular constant or an abstract variable.
+    fn width(comp: &ir::Component, expr: ir::ExprIdx) -> calyx::Width {
+        match comp.get(expr) {
+            ir::Expr::Param(p) => calyx::Width::Param {
+                value: Compile::param_name(comp, *p).into(),
+            },
+            ir::Expr::Concrete(val) => calyx::Width::Const { value: *val },
+            ir::Expr::Bin { .. } | ir::Expr::Fn { .. } => {
+                panic!("Port width must be a parameter or constant.")
```
rachitnigam, line 73: Use `comp.internal_error` instead of `panic!`

src/ir_passes/compile.rs
```diff
@@ -0,0 +1,754 @@
...
+            }
+        }
+    }
+
+    /// Compiles an [ir::ExprIdx] into a [u64].
+    /// Expects the [ir::ExprIdx] to be a single constant value
+    fn u64(comp: &ir::Component, expr: ir::ExprIdx) -> u64 {
```
rachitnigam, line 80: This is already a method on `ExprIdx`

src/ir_passes/compile.rs
```diff
@@ -0,0 +1,754 @@
...
+                        direction: calyx::Direction::Input,
+                        attributes: vec![(
+                            calyx::Attribute::Unknown("fil_event".into()),
+                            1,
+                        )]
+                        .try_into()
+                        .unwrap(),
+                    }),
+            )
+            .collect();
```
rachitnigam, line 133: This chain is quite complex. Let's break it up to make it more readable

src/ir_passes/compile.rs
```diff
@@ -0,0 +1,754 @@
...
+                }
+            }))
+            .chain(
+                comp.events()
+                    .idx_iter()
+                    .filter_map(|idx| Compile::interface_port_name(comp, idx))
+                    .into_iter()
+                    .map(|name| calyx::PortDef {
+                        name: name.into(),
+                        width: wfu(1),
```
rachitnigam, line 123: The function name `wfu` is not very descriptive.

src/ir_passes/compile.rs
```diff
@@ -0,0 +1,754 @@
...
+        for pd in &mut ports {
+            if let Some(pair) = INTERFACE_PORTS
+                .iter()
+                .find(|(_, (n, _, _))| *n == pd.name.as_ref())
+            {
+                assert!(
+                    pair.1 .2 == pd.direction,
+                    "Expected {} to be an {:?} port, got {:?} port.",
+                    pair.1 .0,
+                    pair.1 .2,
```
rachitnigam, line 148: Destructure the pair in the `if-let` above

src/ir_passes/compile.rs
```diff
@@ -0,0 +1,754 @@
...
+                    pd.direction
+                );
+                // TODO: Assert width equality
+                interface_ports.remove(pair);
```
rachitnigam, line 152: This `TODO` item needs more details if we're going to merge the changes with TODO. Alternatively, address the TODO

src/ir_passes/compile.rs
```diff
@@ -0,0 +1,754 @@
...
+                pd.attributes.insert(pair.0 .0, pair.0 .1);
+            }
+        }
+
+        if is_comp {
+            // add remaining interface ports if not found
+            for (attr, (name, width, dir)) in interface_ports {
```
rachitnigam, line 159: Where is the check to see if one of these ports is already present?

src/ir_passes/compile.rs
```diff
@@ -0,0 +1,754 @@
...
+    fn interface_port_name(
+        comp: &ir::Component,
+        idx: ir::EventIdx,
+    ) -> Option<String> {
+        if comp.get(idx).has_interface {
+            Some(if let Some(src) = &comp.src_info {
+                src.interface_ports.get(&idx).unwrap().to_string()
+            } else {
+                format!("ev{}", idx.get())
+            })
```
rachitnigam, line 220: This is a hard to read nesting of things. Use methods like `map` and `unwrap_or_else` on option to make it more readable

src/ir_passes/compile.rs
```diff
@@ -0,0 +1,754 @@
...
+        } else {
+            None
```
rachitnigam, line 222: This should just be moved to the top so we don't have to nest:
```
if !comp.get(...) {
  return None;
}
// rest of the code
```

rachitnigam (changes requested):
Got partially through the compile.rs file. It is getting quite complex. I recommend creating `ir_passes/lower/` folder and then separating some of the logically distinct things into separate files with well defined interfaces.

Other comments are about style and making use of already defined methods and defining reusable methods. Furthermore, something as complex as lowering needs to have more comments and readable code so other people can also maintain it in the future without much pain.

src/ir/from_ast/astconv.rs
UnsignedByte: We need to mutate src_info if we are adding info to it.

src/ir/from_ast/astconv.rs
UnsignedByte: I can't because it has to borrow comp as immutable, which is also needed as mutable further into the statement.

src/ir_passes/compile.rs
UnsignedByte: How can I also have a name for the full tuple so I can use it as a key for the hashmap without having to reconstruct it?

src/ir_passes/compile.rs
UnsignedByte: in line 149 the present interface ports are already removed from the hashmap.

src/ir/comp.rs
UnsignedByte: Should we directly put unannotated ports into the component instead?

src/ir/utils.rs
rachitnigam: This was incorrectly marked as "resolved"

src/ir_passes/lower/build_ctx.rs
```diff
@@ -0,0 +1,347 @@
+use std::{collections::HashMap, iter, rc::Rc};
+
+use crate::ir::{self, Ctx};
+use calyx::structure;
+use calyx_ir::{self as calyx, RRC};
+use itertools::Itertools;
+
+use super::utils as compile_utils;
+use super::Fsm;
```
rachitnigam, line 9: No spaces between imports

src/ir_passes/lower/build_ctx.rs
```diff
@@ -0,0 +1,347 @@
...
+    pub builder: calyx::Builder<'a>,
+    pub binding: &'a mut Binding,
+    pub comp: &'a ir::Component,
+    ctx: &'a ir::Context,
+    lib: &'a calyx::LibrarySignatures,
+    fsms: HashMap<ir::EventIdx, Fsm>,
+    /// Mapping from [ir::InstIdx]s to the calyx cell instantiated.
+    instances: HashMap<ir::InstIdx, RRC<calyx::Cell>>,
+    /// Mapping from [ir::InstIdx]s to a reference of the calyx cell instantiated/invoked
+    invokes: HashMap<ir::InvIdx, RRC<calyx::Cell>>,
```
rachitnigam, line 56: If all instances and all invokes will have a binding here, we should instead use `DenseIndexInfo` instead

src/ir_passes/lower/compile.rs
```diff
@@ -0,0 +1,300 @@
...
+
+use crate::ir::{self, Ctx, Traversal};
+use calyx_frontend as frontend;
+use calyx_ir as calyx;
+use calyx_utils::CalyxResult;
+
+use super::{
+    build_ctx::{Binding, BuildCtx},
+    utils::INTERFACE_PORTS,
+};
```
rachitnigam, line 16: No spaces between imports

src/ir_passes/lower/compile.rs
```diff
@@ -0,0 +1,300 @@
...
+                assert!(
+                    dir == &pd.direction,
+                    "Expected {} to be an {:?} port, got {:?} port.",
+                    name,
+                    dir,
+                    pd.direction
+                );
+                // TODO: should also assert that the width of the matching port is the same as what we expect
+                // We'd also need an equality function on `CW` types to do this, which we don't have at the moment
+                // for [calyx::Width].
```
rachitnigam, line 121: Oh, maybe we should open an issue for the Calyx repo?

src/ir_passes/lower/compile.rs
```diff
@@ -0,0 +1,300 @@
...
+            log::debug!("Defining main component {idx}");
+            component.attributes.insert(calyx::BoolAttr::TopLevel, 1);
+        }
+
+        let builder = calyx::Builder::new(&mut component, lib).not_generated();
+        let mut ctx = BuildCtx::new(ctx, idx, bind, builder, lib);
+
+        // Calculate the max states needed for each FSM for every event.
+        // Done by finding the furthest offset referenced in any [Time] in the component.
+        let mut max_states = HashMap::new();
```
rachitnigam, line 207: The max states computation needs to be moved out somewhere (maybe as a method on `Component`) because it is also used to generate `INTERFACE.json` when we're running our test harness. Currently everything works out because `fud/filament.py` doesn't use `--ir` with `--dump-interface` but it will break when we remove the ast-based compilation.

We can leave it here for now and extract in a following PR but please open an issue.

src/ir_passes/lower/utils.rs
```diff
@@ -0,0 +1,104 @@
...
+        Some(
+            comp.src_info
+                .as_ref()
+                .map(|src| src.interface_ports.get(&self).unwrap().to_string())
+                .unwrap_or_else(|| format!("ev{}", self.get())),
+        )
+    }
+}
+
+impl ir::ExprIdx {
```
rachitnigam, line 34: I'm not a huge fan of adding new `impl` on a type defined in another folder. In the future, we'd want `filament-ir` to be a separate crate which will break this.

src/ir_passes/lower/utils.rs
```diff
@@ -0,0 +1,104 @@
...
+            ir::Expr::Concrete(val) => calyx::Width::Const { value: *val },
+            ir::Expr::Bin { .. } | ir::Expr::Fn { .. } => comp
+                .internal_error("Port width must be a parameter or constant."),
+        }
+    }
+}
+
+impl ir::ParamIdx {
+    /// Returns the name of this parameter.
+    pub fn name(self, comp: &ir::Component) -> String {
```
rachitnigam, line 59: Maybe this should be a method defined on `src_info` struct instead?

src/ir_passes/lower/utils.rs
```diff
@@ -0,0 +1,104 @@
...
+        comp.src_info
+            .as_ref()
+            .map(|src| src.params.get(&self).unwrap().to_string())
+            .unwrap_or_else(|| format!("pr{}", self.get()))
+    }
+}
+
+impl ir::PortIdx {
```
rachitnigam, line 67: Same comment about not defining `ir` `impl`s here

rachitnigam (approved):
There are a couple of final style comments and the problem of unnannotated ports that I left comments for. I'm going to approve this PR so you can merge once those things are done. Good job on this! Compilation is arduous!

src/ir_passes/lower/utils.rs
UnsignedByte: The main thing is it needs to generate a default name as well if src_info is none, and furthermore implementing it in `ir` on the idxs or on compile seems a little unnecessary as these functions will only ever be used in compile.

src/ir_passes/lower/utils.rs
UnsignedByte: Should we just define a helper function that takes `ir::ExprIdx` instead then?

src/ir_passes/lower/build_ctx.rs
UnsignedByte: Instances and invokes that are deleted will not be in this mapping.

src/ir_passes/lower/compile.rs
UnsignedByte: Will do

src/ir_passes/lower/utils.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub fn name(self, ctx: &impl Ctx<ir::Component>) -> String {
+        ctx.get(self)
+            .src_info
+            .as_ref()
+            .map(|src| src.name.to_string())
+            .unwrap_or_else(|| format!("comp{}", self.get()))
+    }
+}
+
+impl From<&ir::Direction> for calyx::Direction {
```
UnsignedByte, line 97: If we are moving ir into its own crate eventually, is there an alternative to implementing `PartialEq` on `ir::Direction`? Should we move this directly to where direction is defined?

UnsignedByte: Removed this instead as it is only used once.

PR #197, opened 2023-06-30
generator for 2D stencils

rachitnigam:
Generalizing the code for stencil generators. This will allow us to write code for a parametric version of our aetherling inspired `conv2d` kernel


rachitnigam:
Not sure when I'll get a chance to work on this again so closing for now

PR #198, opened 2023-06-30
Fix minor typo

hasheddan:
Fixes a minor typo of falsify.

> Just touring the code base after watching the excellent recent talk from @rachitnigam. Noticed this tiny one and thought I would get my feet wet with the contribution flow :) Excited by the direction of filament and looking forward to contributing more!


rachitnigam:
Woo! All contributions welcome!! Let us know if you're trying to implement a design in Filament or want to hack on something that we can get you started!

PR #206, opened 2023-07-08
Syntax changes for interval syntax

rachitnigam:
Unifies the syntax for availability intervals. The distinction between bundle and port syntax is very simple:
```
bundle foo[N]: [G, G+1] 32
```
and
```
bundle bar: [G, G+1] 32
```

Also note that since we don't use the bundle parameter in the availability interval, we don't have to specify it. Additionally, this difference is consistent with port definitions in the signature as well.


rachitnigam:
We should try to merge this when all the tests are working with the IR flow

Issue #199, opened 2023-07-02
Floating point library bug somewhere


UnsignedByte:
The floating point module has a rare bug leading to different behavior from expectation. This could either be due to the 5 shift checks in the other 3 libraries or some other reason.

Full trace (copied from a run of runt in workflow) shown [here](https://github.com/cucapra/filament/actions/runs/5436020310/jobs/9885576606#step:7).

rachitnigam:
GitHub actions logs are not retained so let’s get the minimal failing run in this issue.

rachitnigam:
@UnsignedByte I'm going to disable the testing of your FP library since its broken and it's stopping other PRs from being merged. Please fix it soon and we can re-enable it.

rachitnigam:
Here is the minimal reproducible example. The data file is:
```json
{
    "left": [
        479189982
    ],
    "right": [
        2626205011
    ]
}
```

And the command to run (after saving the file to `data.json`) is:
```sh
fud e --to cocotb-out -s cocotb.data data.json -s futil.flags ' -d canonicalize' evaluation/floating-point/add/harness.fil | jq
```

The generated output is:
```json
{
  "out": {
    "0": [
      478267811
    ]
  },
  "verilog_pipe": {
    "0": [
      478267811
    ]
  },
  "verilog_nopipe": {
    "0": [
      478267811
    ]
  },
  "filament_lib": {
    "0": [
      442814848
    ]
  },
  "cycles": 5
}
```

rachitnigam:
I'm not sure if we should take the current FP implementation as the source of truth. One possible way to figure out if we're doing things correctly is using something like [TestFloat](http://www.jhauser.us/arithmetic/TestFloat.html) library. Checking the correctness of floating point in general is a morass of complications so maybe the right approach is using Berkeley's [HardFloat](https://github.com/ucb-bar/berkeley-hardfloat) library and move the correctness claims to them.

UnsignedByte:
> I'm not sure if we should take the current FP implementation as the source of truth. One possible way to figure out if we're doing things correctly is using something like [TestFloat](http://www.jhauser.us/arithmetic/TestFloat.html) library. Checking the correctness of floating point in general is a morass of complications so maybe the right approach is using Berkeley's [HardFloat](https://github.com/ucb-bar/berkeley-hardfloat) library and move the correctness claims to them.

Confirmation from C++: the current FP implementations are incorrect, the filament library matches a C++ implementing `iec559` floats.

[...]

PR #207, opened 2023-07-08
disable FP lib testing for FP add

rachitnigam:


rachitnigam:
Disabling testing of #199 because of the bug in the implementation of FP add. Can re-enable it once we fix it.

PR #208, opened 2023-07-08
Invoke and Instance info

UnsignedByte:
Small change to add invoke info and instance info to the IR (previously untracked).


src/ir/comp.rs
```diff
@@ -343,3 +343,9 @@ impl MutCtx<Instance> for Component {
         self.instances.get_mut(idx)
     }
 }
+
+impl MutCtx<Info> for Component {
```
rachitnigam, line 347: Why do we have to mutate `Info`? My thought for not exposing this method was because once an piece of information is created, it shouldn't really be modified because we no longer remember where it came from

src/ir/from_ast/astconv.rs
```diff
@@ -834,6 +851,109 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
+            comp.events()
+                .iter()
+                .filter_map(|(idx, event)| {
+                    if let ir::EventOwner::Sig { .. } = event.owner {
+                        if let ir::Info::Event { name, .. } =
+                            comp.get(event.info)
+                        {
+                            Some((*name, idx))
+                        } else {
+                            unreachable!("Incorrect info type for event")
```
rachitnigam, line 896: Use `let`-`else` here so that we don't nest things

src/ir/from_ast/astconv.rs
```diff
@@ -834,6 +851,109 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
+                })
+                .collect(),
+        );
+        // Adds all signature events into the mapping by name
+        event_map.push(
+            idx,
+            comp.events()
+                .iter()
+                .filter_map(|(idx, event)| {
+                    if let ir::EventOwner::Sig { .. } = event.owner {
```
rachitnigam, line 890: This can also be a `let`-`else` with the `else` branch just doing `return None`

src/ir/from_ast/astconv.rs
```diff
@@ -834,6 +851,109 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
+                .ports
+                .iter()
+                .map(|idx| {
+                    let port = comp.get(*idx);
+                    // gets the other port from the other component and inserts into port info
+                    if let ir::Info::Port { name, .. } = comp.get(port.info) {
+                        let inst = comp.get(inv.inst);
+                        (*idx, *port_map.get(inst.comp).get(name).unwrap())
+                    } else {
+                        unreachable!("Incorrect info type for port")
```
rachitnigam, line 923: Also `let`-`else` here

src/ir/control.rs
```diff
@@ -97,6 +99,10 @@ pub struct Invoke {
     pub inst: InstIdx,
     // The ports defined by this invocation
     pub ports: Vec<PortIdx>,
+    // The events defined by this invocation
+    pub events: Vec<EventIdx>,
```
rachitnigam, line 103: This seems incorrect: Instances don't define events; invocations do. What is this supposed to represent?

src/ir/from_ast/astconv.rs
```diff
@@ -834,6 +851,109 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
     }
 }
 
+/// Build connect info between invokes and components
```
rachitnigam, line 854: This comment should be more detailed. What is "connect info" supposed to track and where is it used?

src/ir/from_ast/astconv.rs
```diff
@@ -855,5 +975,8 @@ pub fn transform(ns: ast::Namespace) -> ir::Context {
         let ir_comp = BuildCtx::comp(comp, idx, &sig_map);
         ctx.comps.checked_add(idx, ir_comp);
     }
+
+    connect(&mut ctx);
```
rachitnigam, line 979: Hmm, I'm a bit confused about this. I thought that this is operating on the `ast::Connect` nodes but it seems to do something completely different. Can you explain what this method is supposed to do and why it is running after the all the components have been transformed?

src/ir/comp.rs
UnsignedByte: The issue is a second pass after all components are created is necessary for this pass to work. We could possibly store an Option<InfoIdx> to a different type of information instead, but the option becomes redundant as it is necessary for all invokes. I also thought about storing a dummy InfoIdx that points to empty, but that requires an extra InfoIdx to be created each time which might be unwanted.

UnsignedByte: The idea is to track the Idxs between the event defined by an invoke + the event defined in the component it corresponds to (and the same with ports). The reason the second pass is necessary is these Idxs are not all generated until all components have been compiled.

rachitnigam: A large part of the AST functionality will be removed once the IR flow becomes the default so we'll need to define something like that anyways. Also, in astconv, when you're defining the event binds, you should have access to the information about which event is connected to which right?

rachitnigam (approved):
LGTM!

PR #210, opened 2023-07-10
Add entrypoint

UnsignedByte:
Adds entrypoint information to the IR.

### DIFF
--- a/src/ir/comp.rs
+++ b/src/ir/comp.rs
@@ -9,6 +9,7 @@ use crate::utils::Idx;
 #[derive(Default)]
 pub struct Context {
     pub comps: IndexStore<Component>,
+    pub entrypoint: Option<CompIdx>,
 }
 
 /// A IR component. If `is_ext` is true then this is an external component.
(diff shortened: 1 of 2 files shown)


src/ir/comp.rs
```diff
@@ -9,6 +9,7 @@ use crate::utils::Idx;
 #[derive(Default)]
 pub struct Context {
     pub comps: IndexStore<Component>,
+    pub entrypoint: Option<CompIdx>,
```
rachitnigam, line 12: This shouldn't be an optional by the time we hit the IR. Instead, if the user did not provide an entry point, we should look for "main" and if it missing, then we should throw an error

rachitnigam (commented):
Comment about removing the `Option` around the `entrypoint`. Fix and merge at your discretion

src/ir/comp.rs
UnsignedByte: I left the Option in case in the future we wanted to have the ability to compile without entrypoint, but I can remove that for now.

UnsignedByte: Further issue: `Default` cannot be derived if the option is not included, as the entrypoint becomes required. Should we remove this trait altogether and just add a new function?

PR #213, opened 2023-07-11
IR Validation

rachitnigam:


src/ir/control.rs
```diff
@@ -97,6 +97,8 @@ pub struct Invoke {
     pub inst: InstIdx,
     // The ports defined by this invocation
     pub ports: Vec<PortIdx>,
+    /// The events the invocation uses
+    pub events: Vec<EventBind>,
```
rachitnigam, line 101: I think @UnsignedByte also added this so it will cause a conflict with one of the PRs when merged. Hopefully won't be too annoying to fix.

rachitnigam, line 101: Oh, this isn't quite right. It should track `EventIdx`, not `EventBind`s since the invoke defines which events are defined (in addition to how they are bound)

gabizon103: Got it, will fix

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+        // Validate commands
+        for cmd in self.comp.cmds.iter() {
+            self.command(cmd.clone());
+        }
+    }
+
+    /// An Expr is valid if:
+    /// (1) It is bound in the component
+    /// (2) All the exprs it uses are also bound
+    pub fn expr(&self, eidx: ir::ExprIdx) {
```
rachitnigam, line 66: This can get a bit expense but here is an invariant in the construction of expressions that we can potentially use: if `%n` is defined, then all expressions `%i` where `i` < `n` are also defined. Furthermore, a compound expression's index (for example `ir::Expr::Bin`) has a higher index than its children because the children must be constructed before the compound expression. This means, we should only have to check if the expression is defined in the and ignore its children if the underlying expression is a compound.

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+            }
+        }
+
+        // check (3)
+        self.expr(*width);
+
+        // let p: for<#i> [G+#N, G+i+1]
+        // let p: for<%pr1> [G+%pr2, G+%pr1+1]
+        // let p0: for<%pr3> [G+%pr2, G+%pr1+1] %pr1 -> %pr3
+        // %pr1: owned by port p
```
rachitnigam, line 156: Unused comment

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+
+        // check (1)
+        match owner {
+            ir::EventOwner::Sig => {
+                /* Can't check because the sig does not contain this info */
+            }
+            ir::EventOwner::Inv { inv: iidx } => {
+                let ir::Invoke {
+                    inst: _,
+                    ports: _,
```
rachitnigam, line 173: Use `ir::Invoke { events, .. }` to ignore the fields.

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+        match owner {
+            ir::ParamOwner::Sig | ir::ParamOwner::Loop => {
+                /* Nothing to check */
+            }
+            ir::ParamOwner::Bundle(port_idx) => {
+                let ir::Port {
+                    owner: _,
+                    width: _,
+                    live: _,
+                    info: _,
```
rachitnigam, line 249: Use `..` to ignore fields

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+    pub fn range(&self, range: ir::Range) {
+        let ir::Range { start, end } = range;
+        self.time(start);
+        self.time(end);
+    }
+
+    /// A param is valid if:
+    /// (1) It is defined in the component
+    /// (2) Its owner is defined in the component
+    /// (3?) Its owner points to it?
```
rachitnigam, line 234: This doesn't seem to be checked in the current implementation

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+            ir::ParamOwner::Sig | ir::ParamOwner::Loop => {
+                /* Nothing to check */
+            }
+            ir::ParamOwner::Bundle(port_idx) => {
+                let ir::Port {
+                    owner: _,
+                    width: _,
+                    live: _,
+                    info: _,
+                } = &self.comp.get(*port_idx); // (2) this will panic if port not defined
```
rachitnigam, line 250: This needs to check that the underlying port points to this param as well

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+            }
+        }
+    }
+
+    /// An invoke is valid if:
+    /// (1) Its ports are defined in the component
+    /// (2) Ports defined by invoke point to it
+    ///     i.  port() checks that the invoke owns the port
+    ///         invoke() checks that the ports an invoke defines are owned by it
```
rachitnigam, line 259: Invokes should also check the events they define

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+    fn invoke(&self, iidx: ir::InvIdx) {
+        let ir::Invoke { inst: _, ports, .. } = &self.comp.get(iidx);
+
+        // check (1) and (2)
+        for pidx in ports {
+            // (1) looking up the port will error if it doesn't exist
+            let ir::Port { owner, .. } = &self.comp.get(*pidx);
```
rachitnigam, line 266: Instead of destructuring, you can just do `.owner`

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+            // check (2)
+            self.expr(*expr);
+        }
+        // check (3) and (4)
+        let comp_params = self
+            .ctx
+            .get(*comp)
+            .params()
+            .iter()
+            .filter(|(_idx, param)| param.is_sig_owned())
```
rachitnigam, line 306: Use `_` instead of `_idx`

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+            index,
+            start,
+            end,
+            body,
+        } = lp;
+        self.param(index);
+        self.expr(start);
+        self.expr(end);
+        for cmd in body.iter() {
+            self.command(cmd.clone());
```
rachitnigam, line 420: Shouldn't be cloning things in this pass. Can it use a reference instead?

src/ir/validate.rs
```diff
@@ -0,0 +1,451 @@
...
+    /// (2) Everything in its then-branch is valid
+    /// (3) Everything in its alt-branch is valid
+    fn if_stmt(&self, if_stmt: ir::If) {
+        let ir::If { cond, then, alt } = if_stmt;
+        self.prop(cond);
+        for cmd in then.iter() {
+            self.command(cmd.clone());
+        }
+        for cmd in alt.iter() {
+            self.command(cmd.clone());
```
rachitnigam, line 435: Cloning bad. A pass that checks things shouldn't have to clone things

rachitnigam (commented):
Good start! Couple of style changes and maybe 1-2 actual changes. Can re-review when those are done.

src/ir/control.rs
gabizon103: Oh Edmund might've done it, but I know last week I also talked to you about this so it might've been me. I sorta implemented this in the mono branch, but when I realized it was wrong (for reasons besides it being an `EventBind` instead of an `EventIdx`) so I implemented it correctly here, so I will just use this version when I merge it

src/ir/validate.rs
rachitnigam: Yeah, kind of tricky but the well-formedness constraint for `IndexStore` is that it doesn't even give you `%2` till you've constructed `%1` and `%0`. This is a constraint that cloning an `Interned` should maintain is more of a property of the intern store than the component.

Think of it this way: the validation pass is responsible for checking relations *between* different tables while each table is responsible for maintaining its own variants. Does that sound somewhat reasonable to you?

PR #214, opened 2023-07-13
Fix floating point addition implementations and use external testing.

UnsignedByte:
Fixes disparity between floating point addition implementations or verifies the correct implementation.


rachitnigam:
It's not clear to me which implementation is correct about the output here. I'm not super inclined to test either the original one or the new one. Long-term, we need to figure out a better testing strategy for floating point (or completely give up on implementing it ourselves and use [HardFloat](https://github.com/ucb-bar/berkeley-hardfloat) instead). 

See [this comment about potential testing strategies](https://github.com/cucapra/filament/issues/199#issuecomment-1627553988).

UnsignedByte:
Closing this because we are probably not going to use hardfloat for the forseeable future, and it is not worth spending so much time and effort on.

PR #216, opened 2023-07-14
New ir mono

rachitnigam:


src/ir/expr.rs
```diff
@@ -14,10 +14,20 @@ pub enum Expr {
     },
     Fn {
         op: ast::UnFn,
-        args: Box<[ExprIdx]>,
+        args: Box<Vec<ExprIdx>>,
     },
 }
 
+impl Expr {
+    pub fn as_concrete(&self) -> Option<u64> {
```
rachitnigam, line 22: There is another `as_concrete` method on `ExprIdx`. In general, prefer defining methods on `ExprIdx` instead of `Expr` because most of the time, user code will have access to the former

src/ir/fact.rs
```diff
@@ -112,6 +112,16 @@ impl fmt::Display for Prop {
     }
 }
 
+impl Prop {
+    pub fn as_concrete(&self) -> Option<bool> {
```
rachitnigam, line 116: Similarly, prefer `PropIdx` instead of `Prop` for defining methods

src/ir/from_ast/astconv.rs
```diff
@@ -163,22 +163,18 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
-                    ast::Op::Add => l.add(r, self.comp),
-                    ast::Op::Mul => l.mul(r, self.comp),
-                    ast::Op::Sub => l.sub(r, self.comp),
-                    ast::Op::Div => l.div(r, self.comp),
-                    ast::Op::Mod => l.rem(r, self.comp),
-                }
+                let lhs = self.expr(*left);
+                let rhs = self.expr(*right);
+                // The .add call simplifies the expression if possible
+                self.comp.add(ir::Expr::Bin { op, lhs, rhs })
```
rachitnigam, line 169: This needs to be changed back because it might cause a stackoverflow right?

src/ir/subst.rs
```diff
@@ -1,3 +1,5 @@
+use itertools::Itertools;
+
```
rachitnigam, line 2: No space between imports

src/ir/subst.rs
```diff
@@ -18,6 +20,10 @@ where
             .iter()
             .find_map(|(k, v)| if k == key { Some(v) } else { None })
     }
+
+    pub fn insert(&mut self, key: K, value: V) {
+        self.0.push((key, value));
```
rachitnigam, line 25: I'm not sure if we should be mutably updating bindings. Or at least, exposing this as the function signature. It might be better to consume the binding by making the function take `self` so that it's explicit that the old binding is now truly gone

src/ir/comp.rs
```diff
@@ -16,6 +18,12 @@ impl Context {
     pub fn is_main(&self, idx: CompIdx) -> bool {
         Some(idx) == self.entrypoint
     }
+
+    /// Add a new component to the context
+    pub fn comp(&mut self, is_ext: bool) -> CompIdx {
+        let comp = Component::new(is_ext);
+        self.comps.add(comp)
```
rachitnigam, line 25: Use `self.add` here instead because we implement `Ctx`

src/ir/expr.rs
```diff
@@ -14,10 +14,20 @@ pub enum Expr {
     },
     Fn {
         op: ast::UnFn,
-        args: Box<[ExprIdx]>,
+        args: Box<Vec<ExprIdx>>,
```
rachitnigam, line 17: No need to `Box` if you have a `Vec` here. The boxing was only needed because you cannot store a slice directly

src/ir_passes/new_mono.rs
```diff
@@ -0,0 +1,868 @@
...
+    /// Times
+    time_map: HashMap<ir::TimeIdx, ir::TimeIdx>,
+    /// Ports
+    port_map: HashMap<ir::PortIdx, ir::PortIdx>,
+    /// Exprs
+    expr_map: HashMap<ir::ExprIdx, ir::ExprIdx>,
+    /// Props
+    prop_map: HashMap<ir::PropIdx, ir::PropIdx>,
+    /// Params
+    param_map: HashMap<ir::ParamIdx, ir::ParamIdx>,
```
rachitnigam, line 46: Should these be `DenseInfoMap` instead of a `HashMap`?

src/ir_passes/new_mono.rs
```diff
@@ -0,0 +1,868 @@
...
+        for (idx, port) in self.underlying.ports().iter() {
+            if port.is_sig_in() || port.is_sig_out() {
+                self.port(idx);
+            }
+        }
+        for cmd in &self.underlying.cmds {
+            let cmd = self.command(cmd);
+            self.base.cmds.extend(cmd);
+        }
+        for (idx, _) in self.underlying.props().iter() {
```
rachitnigam, line 60: You can use `iter_idx` here instead

src/ir_passes/new_mono.rs
```diff
@@ -0,0 +1,868 @@
...
+                new_idx
+            }
+            ir::Expr::Fn { op, args } => {
+                let args = args.iter().map(|idx| self.expr(*idx)).collect_vec();
+                let new_expr = self.base.func(ir::Expr::Fn {
+                    op,
+                    args: Box::new(args),
+                });
+                let new_idx = self.base.add(new_expr);
+                self.expr_map.insert(expr, new_idx);
```
rachitnigam, line 174: Seems like all variants do the `expr_map.insert`. Let's move it out of the match?

src/ir_passes/new_mono.rs
```diff
@@ -0,0 +1,868 @@
...
+        port.owner = mono_owner; // update
+
+        new_port
+    }
+
+    /// Add `self.underlying`'s info to `self.base`. Nothing else needs to be done because all the constructs
+    /// we generate and add to `self.base` map to the same source-level info that they did in `self.underlying`
+    fn info(&mut self, info: &ir::InfoIdx) -> ir::InfoIdx {
+        let info = self.underlying.get(*info);
+        self.base.add(info.clone())
```
rachitnigam, line 270: `Info` is stored in an `IndexStore` I think which doesn't deduplicate things which means this will create a lot of copies of the same data over and over.

src/ir_passes/new_mono.rs
```diff
@@ -0,0 +1,868 @@
...
+            .fold(&mut vec![], |acc, cvec| {
+                acc.extend(cvec);
+                acc
+            })
+            .to_vec();
+
+        match cond {
+            ir::Prop::True => self.base.cmds.extend(then),
+            ir::Prop::False => self.base.cmds.extend(alt),
+            _ => panic!("couldnt resolve prop"),
```
rachitnigam, line 598: Use `self.base.internal_error` instead of `panic!`

src/ir_passes/new_mono.rs
```diff
@@ -0,0 +1,868 @@
...
+    /// (1) The source port is invoke-owned and the destination port is also invoke-owned, or
+    /// (2) The source port is sig-owned and the destination port is invoke-owned
+    /// A connect gets handled now if:
+    /// (1) The source port is invoke-owned and the destination port is sig-owned
+    /// (2) The source port is sig-owned and the destination port is also sig-owned
+    /// The reasoning for this is that we see connects where the destination port is invoke-owned before
+    /// we see the actual invoke, so we have to put them in a queue to be handled when we see the invoke
+    /// If the destination port is not invoke owned, then we can handle it now because we've already created
+    /// the monomorphized ports
+    fn handle_connect(&mut self, con: &ir::Connect) -> Vec<ir::Command> {
```
rachitnigam, line 646: Invokes are stored in a separate table so can you iterate over them before walking over the commands and get rid of this delayed handling?

src/ir_passes/new_mono.rs
```diff
@@ -0,0 +1,868 @@
...
+        let binding = underlying
+            .sig_params()
+            .into_iter()
+            .zip(params)
+            .collect_vec();
+
+        // after take(), idx will point to default component
+        let base = std::mem::take(self.ctx.get_mut(base_idx));
+
+        let mut mono = MonoDeferred {
```
rachitnigam, line 789: This should be abstracted in a `new()` method for the struct

src/ir_passes/new_mono.rs
```diff
@@ -0,0 +1,868 @@
...
+    pass: &'a mut Monomorphize<'pass>,
+
+    /// Mapping of underlying invokes (and how many times we've seen it) to base invokes
+    inv_map: HashMap<(ir::InvIdx, u32), ir::InvIdx>,
+    /// Mapping from underlying invokes to how many times we've seen it so far
+    inv_counter: HashMap<ir::InvIdx, u32>,
+    /// Mapping of underlying instances (and how many times we've seen it) to base instances
+    inst_map: HashMap<(ir::InstIdx, u32), ir::InstIdx>,
+    /// Maapping from underlying instances to how many times we've seen it so far
+    inst_counter: HashMap<ir::InstIdx, u32>,
```
rachitnigam, line 28: I'm not sure I've followed what these counters are used for. I see that we count up but why do we need to do that?

src/ir_passes/new_mono.rs
gabizon103: If you have something like `f := new Foo<G>(in)` the commands are laid out as:
```
inst0
p1 <- p0
inv0
```
so when I'm processing a component I just iterate through the commands, meaning I'll see the connects before the invokes. I would iterate over the invokes first to generate them, but in the case where there is a loop in the commands I don't think this would work, because if you have an invoke in a loop the only thing that is in `underlying` is that singular invoke. So my thought was that as we unroll loops, we generate the invokes and then connect their ports

src/ir/subst.rs
gabizon103: What would this look like?

src/ir/from_ast/astconv.rs
gabizon103: Not as far as I can tell? `add()` doesn't simplify things, I just made separate functions that will simplify an expression

src/ir_passes/new_mono.rs
rachitnigam: I see! We can try to make it an IR invariant that things must get defined before they are used. For example, the `invoke` statement defines the input and output ports:
```
inv, p1 = invoke
p1 <- p0
```
This means, it should come before the port connection in the control program. This is usually called the domination condition where the definition of something must dominate its use.

rachitnigam: This invariant could also be enforced using the validation pass.

src/ir/utils.rs
```diff
@@ -408,9 +408,9 @@ where
     C: Ctx<T>,
 {
     /// A reference to the underlying value.
-    key: Idx<T>,
+    pub key: Idx<T>,
     /// A reference to the owner of the foreign key.
-    owner: Idx<C>,
+    pub owner: Idx<C>,
```
rachitnigam, line 413: Making these `pub` will allow for mutable access to these fields which we want to avoid. I think you can either define a `take()` method that consumes the `Foreign` and returns the two fields or define accessor methods that don't give mutable access to these fields.

src/ir_passes/lower/compile.rs
```diff
@@ -152,7 +152,8 @@ impl Compile {
 
         assert!(
             comp.is_ext,
-            "Attempting to compile non-primitive component as primitive."
+            "Attempting to compile {} non-primitive component as primitive.",
+            idx
```
rachitnigam, line 156:
```suggestion
            "Attempting to compile {idx} non-primitive component as primitive."
```

tests/ir_check/bundle.expect
```diff
@@ -38,9 +38,9 @@ error: source port does not provide value for as long as destination requires
    ┌─ tests/errors/bundle.fil:15:11
    │
 15 │     out = f{#N+1};
-   │     ----  ^^^^^^^ source is available for @[G+1+#N, G+1+1+#N]
+   │     ----  ^^^^^^^ source is available for @[G+#N+1, G+#N+1+1]
```
rachitnigam, line 41: These changes are a bit suspicious. The original order was like this because the `ExprIdx::add` method reorders the arguments based on their underlying indices.

src/ir/utils.rs
UnsignedByte: This is done in bundle elim by the way in case that gets merged.

runt.toml
```diff
@@ -46,7 +46,7 @@ fud e -s cocotb.data {}.data --to cocotb-out {} -s calyx.flags ' -d canonicalize
 # ]
 # expect_dir = "tests/compile/ir/"
 # cmd = """
-# ./target/debug/filament {} --show-ir > /dev/null
+# ./target/debug/filament {} --ir --show-ir > /dev/null
```
rachitnigam, line 49: We should probably enable these tests now?

src/ir/fact.rs
```diff
@@ -130,6 +140,14 @@ impl PropIdx {
         matches!(ctx.get(*self), Prop::False)
     }
 
+    pub fn as_concrete(&self, ctx: &impl Ctx<Prop>) -> Option<bool> {
+        match ctx.get(*self) {
```
rachitnigam, line 144: The same code is duplicated between `Port` and `PortIdx`

src/ir/utils.rs
rachitnigam: Seems like `bundle-elim` no longer does this

src/ir/comp.rs
```diff
@@ -337,6 +362,169 @@ impl Ctx<Invoke> for Component {
...
+                        self.add(Expr::Concrete((arg as f64).log2().ceil() as u64))
+                    }
+                }
+            }
+        }
+    }
+
+    /// Evaluates a binary operation, assuming that all params have been substituted for
+    /// concrete expressions in monomorphization
+    pub fn bin(&mut self, expr: Expr) -> ExprIdx {
```
rachitnigam, line 502: I think the `simplify` pass might have the same code?

src/ir/comp.rs
```diff
@@ -97,14 +103,18 @@ pub struct Component {
     /// Externally facing interface information, used to preserve interface in compilation.
     /// Must be `Some` for toplevel components and externals.
     pub src_info: Option<InterfaceSrc>,
+    /// The file that this component came from.
+    /// Must be `Some` for externals
+    pub filename: Option<String>,
```
rachitnigam, line 108: This information is contained in the context. We probably only need to know whether something is an external component or not which can be exposed as a method on the context.

src/ir/comp.rs
```diff
@@ -328,6 +353,169 @@ impl Ctx<Invoke> for Component {
     }
 }
 
+impl Component {
+    pub fn resolve_prop(&mut self, prop: Prop) -> PropIdx {
```
rachitnigam, line 357: Let's implement this as a Foldable for this logic in a PR after this is merged. Can we open an issue to track that?

src/ir/fact.rs
rachitnigam, line 116: let's remove this method

src/ir/from_ast/astconv.rs
rachitnigam: This is handled by #237

src/ir/printer.rs
```diff
@@ -163,16 +163,16 @@ impl<'a> Printer<'a> {
...
+        // if a.is_port(self.ctx) {
+        //     self.ctx.display(port)
+        // } else {
+        format!(
+            "{}[{}..{})",
+            self.ctx.display(port),
+            self.expr(start),
+            self.expr(end)
+        )
+        //}
```
rachitnigam, line 175: Commented out code here. If this is debugging related, let's move it under `log::debug_enabled!`

src/ir/printer.rs
```diff
@@ -388,7 +388,7 @@ impl<'a> Printer<'a> {
             ir::PortOwner::Inv { dir, .. } => {
                 writeln!(
                     f,
-                    "{:indent$}{}: bundle({dir}) {} {};",
+                    "{:indent$}{} ({idx}): bundle({dir}) {} {};",
```
rachitnigam, line 391: Move under debug printing

src/ir/printer.rs
```diff
@@ -398,7 +398,7 @@ impl<'a> Printer<'a> {
             ir::PortOwner::Local => {
                 writeln!(
                     f,
-                    "{:indent$}{} = bundle {} {};",
+                    "{:indent$}{} ({idx}) = bundle {} {};",
```
rachitnigam, line 401: Move under debug printing

src/ir/validate.rs
```diff
@@ -6,9 +6,9 @@ use super::IndexStore;
 /// The validity condition for each structure is defined in the corresponding method.
 pub struct Validate<'a> {
     /// The component being validated.
-    comp: &'a ir::Component,
+    pub comp: &'a ir::Component,
     /// The context for the program being evaluated
-    ctx: &'a IndexStore<ir::Component>,
+    pub ctx: &'a IndexStore<ir::Component>,
```
rachitnigam, line 11: Why are these public?

src/ir_passes/mono/monodeferred.rs
```diff
@@ -0,0 +1,311 @@
+use super::{monosig::MonoSig, Monomorphize};
+use crate::ir::{self, Ctx};
+use itertools::Itertools;
+
+pub struct MonoDeferred<'a, 'pass: 'a> {
```
rachitnigam, line 5: This should be `pub(super)`

src/ir_passes/mono/monomorphize.rs
```diff
@@ -0,0 +1,163 @@
...
+        self.processed.insert((underlying_idx, params), base_idx);
+
+        let underlying = self.old.get(underlying_idx);
+        let mut mono = MonoDeferred {
+            underlying,
+            pass: self,
+            monosig,
+        };
+
+        mono.gen_comp();
```
rachitnigam, line 123: Make `gen_comp` return the information used below.

src/ir_passes/mono/monosig.rs
```diff
@@ -0,0 +1,753 @@
...
+    /// Mapping of underlying instances (and how many times we've seen it) to base instances
+    pub inst_map: HashMap<(ir::InstIdx, u32), ir::InstIdx>,
+    /// Maapping from underlying instances to how many times we've seen it so far
+    pub inst_counter: HashMap<ir::InstIdx, u32>,
+    /// Hold onto the current PortIdx being handled
+    pub curr_port: Option<ir::PortIdx>,
+    /// Invokes we've already handled, such as when we see an invoke's port before the actual invoke
+    /// so we don't create duplicates
+    pub handled_invokes: Vec<ir::InvIdx>,
+    pub handled_instances: Vec<ir::InstIdx>,
```
rachitnigam, line 32: We discussed making it an IR invariant to have use of ports be dominated by their definition. Can you open an issue to track the creation of a pass that does this?

gabizon103: #245

src/ir/comp.rs
gabizon103: #246

src/ir_passes/mono/monomorphize.rs
rachitnigam: Oh weird okay! Does `gen_comp` take `&mut self` or `self`? The latter should allow returning it but no worries if not!

gabizon103: Yeah it takes `&mut self`

src/ir/comp.rs
gabizon103: The reason I did this I think was because we need to construct a new context, and to do this we need to know what file the underlying component came from so we know the mapping for the new component. I could search through all the entries in the old context's `ext_map` until I find it, but I thought it would be easier to just store that information per component. Let me know which I should do?

rachitnigam: We should keep that information in the `Ctx` and expose a method to get it. This information is probably not going to be used by any pass that acts upon the component and it's only meaningful for externals.

gabizon103: But yeah it does make sense to keep the information in `Ctx`

rachitnigam:
@UnsignedByte looks like assignment check is causing a bunch of failures

PR #217, opened 2023-07-15
Updates the interface port implementation

UnsignedByte:
Changes the representation as discussed with @rachitnigam

### DIFF
--- a/src/ir/from_ast/astconv.rs
+++ b/src/ir/from_ast/astconv.rs
@@ -228,7 +228,7 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
     fn declare_event(
         &mut self,
         eb: &ast::EventBind,
-        interface_port: Option<ir::InfoIdx>,
+        interface_port: Option<(ast::Id, GPosIdx)>,
     ) -> EventIdx {
         let info = self.comp.add(ir::Info::event(
             eb.event.copy(),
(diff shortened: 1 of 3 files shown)


src/ir/from_ast/astconv.rs
```diff
@@ -223,17 +224,23 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
             eb.event.pos(),
             eb.delay.pos(),
+            interface_port,
         ));
         // Add a fake delay of 0.
         let e = ir::Event {
             delay: self.comp.num(0).into(),
             info,
-            interface_port: None,
+            has_interface: true,
```
rachitnigam, line 243: It seems a bit weird to me that `has_interface` is unconditionally true? Shouldn't there some check to ensure that the interface port exists? If that happens at a later point, please add a comment pointing to the function

src/ir/info.rs
```diff
@@ -22,6 +22,7 @@ pub enum Info {
         name: ast::Id,
         bind_loc: GPosIdx,
         delay_loc: GPosIdx,
+        interface_port: Option<super::InfoIdx>,
```
rachitnigam, line 25: Why does this track an `InfoIdx` instead of storing the information inline?

rachitnigam (commented):
Couple of questions inline. Also, can you make sure this can correctly compile the signature of `Register`?

src/ir/info.rs
UnsignedByte: I can change this, but then the fields would be unnamed which might be not as good?

Would we rather have two fields `interface_port_name` and `interface_port_bind_loc` instead, or add a whole new struct?

src/ir/from_ast/astconv.rs
UnsignedByte: Yep, this was a bug I found earlier yesterday and hadn't pushed yet, sorry.

src/ir/info.rs
rachitnigam: Either is fine I think. I've just been adding as many field to the variants because the `Info` object size doesn't matter too much

PR #218, opened 2023-07-15
Add helper functions for connecting invoke ports and events

UnsignedByte:
Adds two helper functions to connect invoke defined ports and eventbinds with the component being invoked's ports and events,

### DIFF
--- a/src/ir/comp.rs
+++ b/src/ir/comp.rs
@@ -30,7 +30,7 @@ impl Context {
         let inv = comp.get(*inv);
         // finds the index of the port in the invocation
         let idx = inv.ports.iter()
-            .filter(|&&p| 
+            .filter(|&&p|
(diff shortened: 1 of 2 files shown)


src/ir/comp.rs
```diff
@@ -18,6 +18,51 @@ impl Context {
...
 
+/// Queries corresponding signature [Event]s and [Port]s for their invoke counterparts
+impl Context {
+    /// Gets the [PortIdx] of the signature port of the invoked component corresponding
+    /// to the given [PortIdx] in the invocation.
+    pub fn get_component_port(
+        &self,
+        comp: &Component,
+        idx: PortIdx,
+    ) -> PortIdx {
```
rachitnigam, line 29: As I mentioned in our sync discussion, this `PortIdx` value is not safe to be used within a component; it is only meaningful for the component that defines the port.

UnsignedByte:
Fixed by #221

PR #221, opened 2023-07-16
Add foreign keys to ports and events

UnsignedByte:


rachitnigam:
Looks like the tests are failing. The error message seems to be from IndexStore or something I think

UnsignedByte:
> Looks like the tests are failing. The error message seems to be from IndexStore or something I think

Seems to be a traversal ordering issue, as this errors sometimes but not always.

rachitnigam:
Indeed!

rachitnigam:
LGTM!

PR #223, opened 2023-07-19
Fix foreign keys compilation order bug.

UnsignedByte:
Declares all components in the context in code-order before mutating them instead of declaring and converting the components in traversal order. This is because rust's `TopologicalSort` package internally uses `HashMap`, meaning the resultant post-order traversal does not have a definite order at runtime, which leads to errors that are
* Not printed in order of position in code
* Not always in the same order between runs

which lead to tests on errors not always passing. This PR re-orders all the components in the context in the definite given order before mutating afterward, which solves this problem.

### DIFF
--- a/src/ast/component.rs
+++ b/src/ast/component.rs
@@ -2,6 +2,7 @@ use super::{Command, Fsm, Id, Invoke, Signature};
 use crate::errors::{Error, FilamentResult};
 use std::fmt::Display;
 
+#[derive(Default)]
 /// A component in Filament
 pub struct Component {
     // Signature of this component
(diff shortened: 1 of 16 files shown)


UnsignedByte:
> Can you runt a couple of times and confirm this works?

I have and as far as I know it should work? I'm a little worried its a compile-time generated issue and not a runtime thing tho (I.E. compiling different code could result in different ordreing, but not running the same compiled build multiple times.)

UnsignedByte:
> > Can you runt a couple of times and confirm this works?
> 
> I have and as far as I know it should work? I'm a little worried its a compile-time generated issue and not a runtime thing tho (I.E. compiling different code could result in different ordreing, but not running the same compiled build multiple times.)

Nevermind, confirmed this isn't the case (differing behavior between runs exists), so the bug should be fixed!

rachitnigam:
Awesome, this looks good to merge!

PR #231, opened 2023-07-22
[NFC] Some IR cleanup

rachitnigam:
Mostly makes the printer use the `Info` data to print out components so that they better match their source-level representation and removes `std::fmt` implementations from IR nodes. The IR printer should be the only way to print meaningful values from the IR.


rachitnigam:
@gabizon103 this changes `invoke` handling in astconv so that `invoke` will appear before the port connections that use ports that the `invoke` defines.

PR #232, opened 2023-07-25
Bundle elimination pass

UnsignedByte:
Creates the bundle elimination pass that removes bundles on a monomorphized program.

For a bundle of length `l`, creates `l` single-element ports to replace it. Also compiles invokes and connects within components to instead use these new ports. 

Modifies the signature of components.


src/ir/expr.rs
```diff
@@ -47,6 +47,41 @@ impl ExprIdx {
         }
     }
 
+    pub fn resolve(
+        self,
+        ctx: &mut impl Ctx<Expr>,
+        binding: &HashMap<ParamIdx, u64>,
+    ) -> ExprIdx {
```
rachitnigam, line 54: I think @gabizon103's mono PR also adds something like this? Also, have you looked at `utils::Subst`? That structure provides an `apply()` method to do this kind of resolving for you. Let me know if you have trouble understanding how it's used---it's kind of tricky.

src/ir/utils.rs
```diff
@@ -408,9 +413,9 @@ where
     C: Ctx<T>,
 {
     /// A reference to the underlying value.
-    key: Idx<T>,
+    pub key: Idx<T>,
     /// A reference to the owner of the foreign key.
-    owner: Idx<C>,
+    pub owner: Idx<C>,
```
rachitnigam, line 418: Don't make these `pub`; that allows someone to mutate them. Instead, let's provide accessor methods.

src/ir_passes/bundle_elim.rs
```diff
@@ -0,0 +1,233 @@
...
+use itertools::Itertools;
+
+use crate::ir::{
+    Access, Command, CompIdx, Component, Connect, Context, Ctx, DenseIndexInfo,
+    Expr, Foreign, InvIdx, Invoke, Liveness, MutCtx, Port, PortIdx, Range,
+    Time,
+};
+
+#[derive(Default)]
+pub struct BundleElim {
```
rachitnigam, line 12: Documentation?

src/ir_passes/bundle_elim.rs
```diff
@@ -0,0 +1,233 @@
...
+            live,
+            info,
+        } = comp.get(pidx).clone();
+
+        let Liveness { idx, len, range } = live;
+
+        let start = comp.get(range.start).clone();
+        let end = comp.get(range.end).clone();
+
+        let len = len.as_concrete(comp).unwrap();
```
rachitnigam, line 31: The length of a port should be a method on the `ir::PortIdx` object since it seems like an extremely common query.

src/ir_passes/bundle_elim.rs
```diff
@@ -0,0 +1,233 @@
...
+                "Bundle ports in the signature are not supported."
+            );
+
+            // need to preserve the original portidx here to save the source information.
+            return vec![pidx];
+        }
+
+        let ports = (0..len)
+            .map(|i| {
+                let binding: HashMap<_, _> = [(idx, i)].into();
```
rachitnigam, line 48: This is where the `Bind` and `Subst` structures come in handy: you don't have to allocate expensive HashMaps.

src/ir_passes/bundle_elim.rs
```diff
@@ -0,0 +1,233 @@
...
+                    offset,
+                });
+                let live = Liveness {
+                    idx,
+                    len: one,
+                    range: Range { start, end },
+                };
+
+                let owner = match &owner {
+                    crate::ir::PortOwner::Sig { dir } => {
```
rachitnigam, line 68: Don't do `crate::` imports. Use `ir::` instead

src/ir_passes/bundle_elim.rs
```diff
@@ -0,0 +1,233 @@
...
+                    owner,
+                    info,
+                    width,
+                })
+            })
+            .collect();
+        // delete the original port
+        comp.delete(pidx);
+        // delete the corresponding parameter
+        comp.delete(idx);
```
rachitnigam, line 99: Looks good! We should also change the `Validate` pass so that it checks that invalid ports etc. are not being used by anyone, either inside the component or through a `Foriegn`.

src/ir_passes/bundle_elim.rs
```diff
@@ -0,0 +1,233 @@
...
+        ports
+    }
+
+    /// Compiles the signature of a component and adds the new ports to the context mapping.
+    fn sig(&self, comp: &mut Component) -> HashMap<PortIdx, Vec<PortIdx>> {
+        // need to collect here because of ownership issues.
+        comp.ports()
+            .idx_iter()
+            .filter(|p| comp.get(*p).is_sig())
+            .collect_vec() // collect here to avoid ownership issues
```
rachitnigam, line 109: What ownership issue occurs if you don't do this? Is it because `port` takes mutable ownership of the component?

src/ir_passes/bundle_elim.rs
```diff
@@ -0,0 +1,233 @@
...
+        ctx: &mut Context,
+    ) -> Vec<Command> {
+        let comp = ctx.get_mut(cidx);
+        let mapping = self.context.get(cidx);
+
+        let Connect { src, dst, info } = connect;
+
+        let start = src.start.as_concrete(comp).unwrap() as usize;
+        let end = src.end.as_concrete(comp).unwrap() as usize;
+        let src = mapping.get(&src.port).unwrap()[start..end].to_vec();
```
rachitnigam, line 149: This code is repeated. Is there a way to abstract it?

src/ir_passes/bundle_elim.rs
```diff
@@ -0,0 +1,233 @@
...
+            .map(|(src, dst)| {
+                (Access::port(src, comp), Access::port(dst, comp))
+            })
+            .map(|(src, dst)| {
+                Command::Connect(Connect {
+                    src,
+                    dst,
+                    info: *info,
+                })
+            })
```
rachitnigam, line 172: Merge the two `map` calls here?

src/ir_passes/bundle_elim.rs
```diff
@@ -0,0 +1,233 @@
...
+            self.context.get_mut(cidx).insert(idx, pl);
+        }
+
+        let cmds = comp
+            .cmds
+            .drain(..)
+            .collect_vec() // collect here to avoid ownership issues
+            .into_iter()
+            .flat_map(|cmd| match cmd {
+                crate::ir::Command::Connect(con) => {
```
rachitnigam, line 203: `crate::` import problem again

src/ir_passes/bundle_elim.rs
```diff
@@ -0,0 +1,233 @@
...
+            .collect()
+    }
+
+    /// Compiles the body of a component and replaces all ports with their expanded versions.
+    fn comp(&mut self, cidx: CompIdx, ctx: &mut Context) {
+        let comp = ctx.get_mut(cidx);
+        // compile invocations
+        for idx in comp.invocations().idx_iter() {
+            let pl = self.inv(idx, comp);
+            self.context.get_mut(cidx).extend(pl);
```
rachitnigam, line 182: Should we be deleting old `invoke`s here? The `invoke` call does not. Alternatively, we can write a "dead code elimination" pass in that can just remove all unused things (ports, invokes, instances) from a component.

UnsignedByte: This mutates the invoke, and doesn't generate new ones, so nothing is created/deleted.

src/ir_passes/bundle_elim.rs
UnsignedByte: Not doing this here because the function requires every part of the deconstructed port, not just the length. If we find the helper function useful elsewhere we can probably add it there.

rachitnigam: That seems fine right? If you call Len on the port indices the port information should be in the cache anyways for the destructuring?

rachitnigam (approved):
LGMT! Resolve last comment at your discretion and merge! We should make the validate changes soon after this to make sure nothing is breaking.

oh also, did this enable any test cases? Just to make sure things are actually working?

UnsignedByte:
Also on the topic of validation, this currently (knowingly) breaks validation rules by deleting a parameter assigned to bundles that don't need a parameter. I was wondering if it might be helpful in the future to make this parameter optional?

PR #235, opened 2023-07-26
Change bundle elimination to use existing foldables and optimize foldable reduction.

UnsignedByte:
Small fix.


rachitnigam:
Looks like this needs to be merged with `main`.

PR #238, opened 2023-07-26
Local port elimination

UnsignedByte:
Fixes bundle elim's local port elimination to prevent local ports from being generated in the ir at all.


src/ir_passes/bundle_elim.rs
```diff
@@ -59,6 +56,9 @@ impl BundleElim {
             return vec![pidx];
         }
 
+        // creates an empty info struct for these new ports
+        let info = comp.add(Info::Empty);
```
rachitnigam, line 60: This will start crashing the `ir::Printer` I think because we do the whole:
```
let ir::Info:::Event { ... } = self.get(info) else {
  unreachable!(...)
};
```

I mentioned in the meeting yesterday that we should come up with a better way of extracting `info` objects, either by storing `Option<InfoIdx>` where we store `InfoIdx` or exposing a bunch of methods on `InfoIdx` like `as_event_info()` that return `Option<(...)>` or something.

Also worth looking into the [`struct_variant`](https://docs.rs/struct-variant/latest/struct_variant/) to make it easy to define a diagonstics struct an add it to the `Info` enum.

src/ir_passes/local_port_elim.rs
```diff
@@ -0,0 +1,63 @@
...
+        for (src, dst) in comp.cmds.iter().filter_map(|cmd| match cmd {
+            Command::Connect(Connect { src, dst, .. }) => {
+                Some((src.port, dst.port))
+            }
+            _ => None,
+        }) {
+            if comp.get(dst).is_local() {
+                self.port_map.insert(dst, src);
+            }
+        }
```
rachitnigam, line 26: These iterators can be replaced with `for_each` instead. Generally, mixing `for` and iterators makes the code pretty unreadable.

src/ir_passes/local_port_elim.rs
```diff
@@ -0,0 +1,63 @@
+use std::collections::HashMap;
+
+use crate::{
+    ir::{Access, Command, Component, Connect, Ctx, Info, PortIdx},
+    ir_visitor::{Action, Visitor},
+};
+
+/// Eliminates local ports from the IR.
+#[derive(Default)]
+pub struct LocalPortElim {
```
rachitnigam, line 10: The existing flow also checks to make sure that all the local ports have exactly one write to them. This means they error if there are multiple writes to the same port location. We should recreate that behavior.

Also, does this assume that all local ports are exactly one element long? That assumption is not true right?

rachitnigam (commented):
Couple of comments but the big ones are:
1. Ensuring that all port indices are assigned to exactly once (one writer and only one writer)
2. Support bundles with sizes greater than 1. No pass before this makes sure that all bundles have size=1

src/ir_passes/local_port_elim.rs
UnsignedByte: It is true because this happens post bundle-elim

src/ir_passes/bundle_elim.rs
UnsignedByte: I can work on this probably.

src/ir_passes/local_port_elim.rs
UnsignedByte: > The existing flow also checks to make sure that all the local ports have exactly one write to them. This means they error if there are multiple writes to the same port location. We should recreate that behavior.

Doesn't bind check or whatever in the first place already check that all ports are only assigned to once? I feel like this would have be an issue that appears before we get to local port elim.

rachitnigam: Nope, it doesnt

rachitnigam: So bundle elim does something other than simplify the signatures of components?

UnsignedByte: Yep, it eliminates bundles throughout the entire ir at the moment, was I mistaken in how that was meant to work?

rachitnigam: Well, that blows up the encoding for local bundles unnecessarily. We need to eliminate signature bundles because we need to explicate the writes to input ports but local ports can just track writes to particular indices.

UnsignedByte: Yeah it does as it's name says: it eliminates all bundles from the code.

rachitnigam:
Can't review this till #239 is merged. In general, you should rebase on `main` when creating a PR that is not related to another one.

src/ir_passes/assignment_check.rs
```diff
@@ -0,0 +1,110 @@
...
+        termcolor::{ColorChoice, StandardStream},
+    },
+};
+use itertools::Itertools;
+
+use crate::{
+    ir::{Component, Connect, Ctx, DisplayCtx, PortIdx},
+    ir_visitor::{Action, Visitor},
+    utils::{GPosIdx, GlobalPositionTable},
+};
```
rachitnigam, line 16: No spaces between imports

src/ir_passes/bundle_elim.rs
```diff
@@ -25,20 +26,28 @@ impl BundleElim {
...
+                let mut group = (*port, idx);
+                let (port, idx) = loop {
+                    match self.local_map.get(&group) {
+                        Some(&g) => group = g,
+                        None => break group,
+                    }
+                };
+                self.context[cidx][&port][idx]
+            })
+            .collect()
```
rachitnigam, line 43: Hm, this is hard to read: In general, mixing iterators and imperative operators like `break` becomes confusing quickly.

src/main.rs
```diff
@@ -46,14 +46,14 @@ fn run(opts: &cmdline::Opts) -> Result<(), u64> {
...
-        }
         ir_passes::Discharge::do_pass(opts, &mut ir)?;
-        ir_passes::BundleElim::do_pass(&mut ir);
         if opts.show_ir {
             ir::Printer::context(&ir, &mut std::io::stdout()).unwrap();
         }
+        ir_passes::AssignCheck::do_pass(opts, &mut ir)?;
+        ir_passes::BundleElim::do_pass(&mut ir);
+        ir_passes::AssignCheck::do_pass(opts, &mut ir)?;
+        // ir_passes::LocalPortElim::do_pass(opts, &mut ir)?;
```
rachitnigam, line 56: Local port elim is no longer executed. Is that intentional?

src/ir_passes/bundle_elim.rs
```diff
@@ -217,20 +217,52 @@ impl BundleElim {
...
+            .filter_map(|cmd| {
+                let Command::Connect(con) = cmd else { return None };
+                let Connect { src, dst, .. } = con;
+
+                // need to check validity here because already deleted ports are still in the connect commands
+                if comp.ports().is_valid(dst.port)
+                    && comp.get(dst.port).is_local()
+                {
+                    let dst_start =
+                        dst.start.as_concrete(comp).unwrap() as usize;
```
rachitnigam, line 233: Can we expose a method on `ExprIdx` to do this in one go? Maybe a `Into` implementation so we could write:
```
let dst_start: usize = dst.start.try_into()
```
One problem is that the `Into` implementation could panic but that can already happen here?

src/ir_passes/bundle_elim.rs
```diff
@@ -217,20 +217,52 @@ impl BundleElim {
...
+                    let dst_end = dst.end.as_concrete(comp).unwrap() as usize;
+                    let src_start =
+                        src.start.as_concrete(comp).unwrap() as usize;
+                    let src_end = src.end.as_concrete(comp).unwrap() as usize;
+                    assert!(
+                        dst_end - dst_start == src_end - src_start,
+                        "Mismatched access lengths for connect `{}`",
+                        Printer::new(comp).connect_str(con)
```
rachitnigam, line 241: Should this be a `comp.internal_error` instead?

src/ir_passes/bundle_elim.rs
UnsignedByte: Can't write an into because it requires a context to resolve, the only thing that would be different from `as_concrete` would be the cast to `usize` which I'm not sure is that useful.

src/ir_passes/bundle_elim.rs
UnsignedByte: Is there a better method to this? I have to mix them because I want to collect a vec of ports at the end, and using `for` instead would require creating a mutable vec which I'm not sure is any cleaner.

src/ir_passes/assignment_check.rs
```diff
@@ -0,0 +1,110 @@
...
+        for ((port, idx), connects) in self.ports.drain() {
+            let l = connects.len();
+            // assigned only once, no problems
+            if l <= 1 {
+                continue;
+            }
+
+            let filtered_cons = connects.into_iter().flatten().collect_vec();
+
+            let missing_info = l - filtered_cons.len();
```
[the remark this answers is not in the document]
UnsignedByte: Is there a reason why we should do it in another way? I feel like this method seems the cleanest to me. I could make it emit the diagnostics immediately upon generation instead to eliminate the second pass if that helps.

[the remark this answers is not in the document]
rachitnigam: I guess I'll leave it to your discretion on this. We should also emit an error if there are no assignments to a particular location of a bundle.

src/ir_passes/bundle_elim.rs
rachitnigam: I usually just bite the bullet and write the imperative version of the code and explicitly allocate a vector using `Vec::with_capacity`

src/ir_passes/bundle_elim.rs
rachitnigam: Gotcha!

rachitnigam:
LGTM! Resolve comments and merge at your discretion!

rachitnigam:
Let's merge this once ready so I can review #240

Issue #233, opened 2023-07-25
IR printer changes


gabizon103:
Comments about things that were removed in #231, but were kind of nice:
1) It was nice to see param owners printed out with params, so I could verify that the ports/params matched up
2) Printing out ports using the source-level names instead of the IR idxs is nice, but when unrolling loops I use the same info as whatever was in the loop body for all the things I generate, so it kind of looks like the same port is being reused when that isn't happening. It would be nice to also have the IR idx somewhere in there.

@rachitnigam if these changes are cool with you, I'll implement them

rachitnigam:
Yeah! One thing we can do is implement this when `log::debug` is enabled (see printing for `ExprIdx`). That way, the default printing mode is "simple" while debug mode prints out more verbose information. Would that work?

rachitnigam:
And yeah, for (2) I was going to say that this might be a good reason *not to copy* the `Info` object since those are semantically distinct ports and shouldn't share a name.

gabizon103:
Yeah that was another question I had - only thing I was iffy about is I'd have to be making up new `ast::Id`s that don't actually exist in the source program

rachitnigam:
Yeah, but technically there shouldn't be any source-level error messages after monomorphization any ways

rachitnigam:
Also, the `Info::Empty` will probably break a bunch of code since we do: 
```let Info::Event { ... } = event.info else { unreachable!(...) }``` 
In a bunch of places in the code.

We should switch to having a different interface for destructuring `Info` in the future so this stops being a problem

UnsignedByte:
#239 will probably affect this change, FYI.

rachitnigam:
@gabizon103 has these been addressed?

gabizon103:
Oops, yeah this was added in #216

PR #239, opened 2023-07-27
Change info struct to use struct variants.

UnsignedByte:
Adds struct variants to the info struct and defines conversion functions that panic when incorrect.

Requested from https://github.com/cucapra/filament/pull/238#discussion_r1276280429

This should probably be merged before #233 is worked on.

### DIFF
--- a/src/ir/info.rs
+++ b/src/ir/info.rs
@@ -3,46 +3,61 @@ use crate::{ast, utils::GPosIdx};
 use codespan_reporting::diagnostic::Diagnostic;
 use struct_variant::struct_variant;
 
+/// An absence of information is still information
 pub struct Empty;
 
+/// Assertion information
 pub struct Assert(pub Reason);
 
+/// For [super::Param]
 pub struct Param {
+    /// Surface-level name of the parameter
     pub name: ast::Id,
     pub bind_loc: GPosIdx,
 }
 
+/// For [super::Event]
 pub struct Event {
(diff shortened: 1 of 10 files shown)


rachitnigam:
Awesome! FWIW, we should expose methods that return `Some(...)` instead of the struct because the `info` is also allowed to be `Info::Empty`

src/ir/info.rs
```diff
@@ -125,19 +132,61 @@ impl Info {
...
             live_loc,
         }
+        .into()
     }
+}
 
-    pub fn unannotated_port(name: ast::Id, width: u64) -> Self {
-        Self::UnannotatedPort { name, width }
-    }
+macro_rules! info_cast {
```
rachitnigam, line 145: Documentation for this macro

src/ir/info.rs
```diff
@@ -1,76 +1,79 @@
...
+    pub dst_loc: GPosIdx,
+    pub src_loc: GPosIdx,
+}
+
+pub struct Port {
+    pub name: ast::Id,
+    pub bind_loc: GPosIdx,
+    pub width_loc: GPosIdx,
+    pub live_loc: GPosIdx,
+}
```
rachitnigam, line 50: This removed all the documentation information from the fields. We should try to preserve those.

rachitnigam (approved):
LGTM! Remaining comments are about missing documentation which can be resolved at your discretion!

Issue #228, opened 2023-07-20
Add mutually recursive test case


UnsignedByte:
We should add a test for mutually recursive components (such as the even-odd problem) because a lot of the IR seems to currently have broken/undefined behavior in this edge case (ref. #226), so we should be testing it.

Issue #170, opened 2023-06-09
[IR] `assert` statements not added to path condition


rachitnigam:
When we write something like this:
```
assert C0;
...
assert C1;
```
Explicit asserts are usually added to the program to make the proof checking easier (for example, the when checking [`GenDeserialize`](https://github.com/cucapra/filament/blob/cb487f662afb613131b0c20feda86f2bd7bd3a5a/primitives/reshape.fil#L35)). However, in the new IR flow, we use a hoisting pass to automatically lift up assertions to the top-level. During this process, we add all the `assume` statements to the path condition for the assertions but not `assert`s. This means that asserts cannot benefit from such proof hints anymore.

The old compiler uses [Activation literals](https://docs.rs/rsmt2/latest/rsmt2/actlit/index.html) which allow the solver to reuse the work when proof relevant constraints. Maybe we should implement them for `easy-smt`.

PR #240, opened 2023-07-31
Astconv fix

### LINKED ISSUE #226: Astconv fails on recursive and mutually recursive components
Astconv currently uses `Traversal`  to generate `Foreign` ports properly, but this method will not work for recursive and mutually recursive components, as they have dependencies on one another. The issue is that the signature ports and events have to be compiled for all dependencies before the commands of the component itself can be compiled, which doesnt work when there are circular dependencies.

An alternative solution is to compile all signature ports and events in a `declare_comp` pass for all components before going through commands (and invokes). This would require declaring and storing all signature information in some mapping before converting each component.

UnsignedByte:
Fixes #226 and adds #228. Compiles signatures into a sigmap before compiling any component bodies.


tests/ir_check/already-assigned.expect
```diff
@@ -0,0 +1,13 @@
+---CODE---
+1
+---STDERR---
+error: Port a{0} assigned multiple times
+   ┌─ tests/errors/already-assigned.fil:10:5
+   │
+10 │     a{0} = a{1};
+   │     ---- assigned here
+11 │     a{0} = a{2};
+   │     ---- assigned here
```
rachitnigam, line 10: Love this error message!

rachitnigam:
Can't review this till #238 is merged since it wasn't rebased on `main`.

UnsignedByte:
> Tests timing out because of #170

Why do the two tests that compile even run? They don't on my machine, do I have the wrong version of runt that needs updating? I purposefully removed their `.expect` files so they wouldn't run.

rachitnigam:
You can rename the test.expect file to test.skip to actually skip the test. By default, runt will run all tests. For tests missing an expect file, it means that we haven’t yet committed a golden output.

rachitnigam:
Still can't review this because it has changes from #238.

src/ir_passes/assignment_check.rs
```diff
@@ -0,0 +1,110 @@
...
+
+use crate::{
+    ir::{Component, Connect, Ctx, DisplayCtx, PortIdx},
+    ir_visitor::{Action, Visitor},
+    utils::{GPosIdx, GlobalPositionTable},
+};
+
+#[derive(Default)]
+/// Makes sure each index in a port is only written to at most once
+/// Must occur after monomorphization.
```
rachitnigam, line 20: This should also ensure that there is at least one write. We need to make sure that there is exactly one write to each location of the bundle.

tests/compile/ir/event-multiple.expect
rachitnigam, line 1: Where are these empty files coming from?

src/ir/from_ast/astconv.rs
```diff
@@ -860,94 +850,79 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
-    // Walk over signatures and build a SigMap
-    for (idx, sig) in ns.signatures().map(|(_, sig)| sig).enumerate() {
-        sig_map.insert(sig.name.copy(), Sig::from((sig, idx)));
-    }
+    // creates an empty context with the main index.
+    let mut ctx = ir::Context {
+        entrypoint: ns
+            .main_idx()
+            // index main component after all externals
+            .map(|idx| crate::utils::Idx::new(ns.externals().count() + idx)),
```
rachitnigam, line 871: Don't use `crate::` namespace inline. Instead, add `use crate::utils`. 

Also, why is this manually creating `Idx` (which is generally a dangerous thing to do). Is it because we need to do a `checked_add` afterwards?

src/ir/from_ast/astconv.rs
```diff
@@ -860,94 +850,79 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
+            builder.comp.cmds = builder.sig(&sig);
 
-    // TODO: Need to handle recursive components as well as mutually recursive components by adding signature ports first in declaration, so that foreign keys can be resolved
-    // declare all components in the proper order
-    for (cidx, comp) in ns.components.iter().enumerate() {
-        let idx = sig_map.get(&comp.sig.name).unwrap().idx;
+            let irsig = Sig::new(idx, &builder.comp, &sig);
+            ((idx, builder, body), (sig.name.take(), irsig))
+        })
+        .unzip();
```
rachitnigam, line 909: This chain of operations is too long and hard to understand. Can we split it up into two before we do the `enumerate` call?

src/ir/from_ast/astconv.rs
```diff
@@ -860,94 +850,79 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
-        // Needs to swap here because we have to own the current component while leaving the context immutable.
-        // TODO: Find a solution here that will still allow us to properly generate Foreign keys in recursive components.
-        std::mem::swap(ctx.get_mut(idx), &mut curr_comp);
-        BuildCtx::comp(&ctx, &mut curr_comp, comp, &sig_map);
-        // swap the component back into place
-        std::mem::swap(ctx.get_mut(idx), &mut curr_comp);
-    }
+    // Compiles and adds all the commands here
+    // Need to do this before adding all the components because each builder borrows the context immutably
+    builders.into_iter().for_each(|(idx, mut builder, cmds)| {
```
rachitnigam, line 918: Can this be a `for` loop instead? Usually, if the iterator chain is just `into_iter().for_each(...)`, I prefer using `for` loops

tests/compile/ir/window-sum.skip
rachitnigam, line 1: This should not be skipped. Let's change the test to not use the `go` signal.

src/ir/from_ast/astconv.rs
UnsignedByte: Yeah the checked add requires it, and it also is required for foreigns to properly work as well, because all the idxs need to be generated before they are actually added to a context.

tests/compile/ir/event-multiple.expect
UnsignedByte: They're the compilation checks for IR, in the future (after mono) we'll add the expected outputs but at the moment it just runs the IR till pre-mono (type checking, etc) to make sure everything can actually be represented properly in the IR.

src/ir_passes/assignment_check.rs
UnsignedByte: Should we make 0 writes a warning or an error? It seems less severe of a problem than multiple writes, but I'm not sure.

rachitnigam: If you read a value from a location but no one writes to it, what is the inlining supposed to do?

I think it should be an error

src/ir/from_ast/astconv.rs
```diff
@@ -254,7 +249,7 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
             ast::TimeSub::Sym { l, r } => {
                 let l = self.time(l);
                 let r = self.time(r);
-                l.sub(r, self.comp)
+                l.sub(r, &mut self.comp)
```
rachitnigam, line 252: Seems like all uses of comp now need a mutable reference to it? We could consider storing a mutable reference to it in the build context or something

src/ir/from_ast/build_ctx.rs
```diff
@@ -59,6 +68,27 @@ impl std::ops::Index<&Id> for SigMap {
     }
 }
 
+impl std::iter::FromIterator<(Id, Sig)> for SigMap {
```
rachitnigam, line 71: I think fromiterator is in the prelude and doesn’t need to be imported from std::iter?

src/ir/from_ast/build_ctx.rs
```diff
@@ -109,16 +137,12 @@ pub(super) struct BuildCtx<'ctx, 'prog> {
...
-        sigs: &'prog SigMap,
-    ) -> Self {
+
+impl<'prog> BuildCtx<'prog> {
+    pub fn new(comp: ir::Component) -> Self {
         Self {
-            ctx,
             comp,
-            sigs,
+            sigs: None,
```
rachitnigam, line 145: Why can this be constructed with no sigs?

src/ir/from_ast/astconv.rs
UnsignedByte: added helper function instead.

src/ir/from_ast/build_ctx.rs
UnsignedByte: Because the sigs need the signatures of the components to be compiled first (to get the foreign portidxs) so there needs to be a period before `builder.sig()` is run where the signatures don't actually exist. We could use SigMap::default() instead which would probably be better.

src/ir/from_ast/astconv.rs
```diff
@@ -845,108 +833,105 @@ impl<'ctx, 'prog> BuildCtx<'ctx, 'prog> {
...
+            // index main component after all externals
+            .map(|idx| Idx::new(ns.externals().count() + idx)),
+        ..Default::default()
+    };
+
+    // Walk over signatures and compile signatures to build a SigMap
+    // Contains a tuple containing three necessary bits of information:
+    // 1. The (optional) name of the component (if it is an external)
+    // 2. The signature of the component
+    // 3. The (optional) body of the component (if it is not an external)
```
rachitnigam, line 877: Awesome, thanks for the detailed comments here!

rachitnigam (approved):
LGTM! Maybe we can add a test for when a bundle location is not written so that we can check the error triggers?

Issue #241, opened 2023-07-31
Remove requirement for DenseIndexInfo to have keys pushed in order


gabizon103:
`DenseIndexInfo::push()` requires things to be pushed in order, which makes it difficult to use in a lot of cases. Instead, we could allow things to be pushed out of order by filling the elements between the last one currently in the store and the one we are trying to push with default values. But then, I guess we would need a way to modify things currently in the store?

PR #242, opened 2023-07-31
Dense index out of order pushes

gabizon103:
#241


gabizon103:
Actually shouldn't we also need a way to update the things that we filled in to be default?

rachitnigam:
Yeah, I think you'd want that interface as well. I was thinking you could have something like `update_at(idx, val)` which pushes a new value at `idx` if nothing is present or updates it if a value is present (with this method subsuming `push_unordered`).

rachitnigam:
The changes look good to me! Let's merge when ready!

PR #237, opened 2023-07-26
Add simplification for expressions upon addition

UnsignedByte:
Moves basic simplification from the `ExprIdx` smart constructors into the `Ctx` `add` function instead.

PR #243, opened 2023-07-31
Simplify propositions upon component addition

UnsignedByte:
Mirrors #237 but for props.

### DIFF
--- a/src/ir/comp.rs
+++ b/src/ir/comp.rs
@@ -1,5 +1,3 @@
-use std::collections::HashMap;
-
 use super::{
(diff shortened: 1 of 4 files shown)


rachitnigam (approved):
LGTM! Any thoughts on what should happen to the simplify code? It is definitely a more expensive form of simplification so maybe we can leave it there.

UnsignedByte:
> LGTM! Any thoughts on what should happen to the simplify code? It is definitely a more expensive form of simplification so maybe we can leave it there.

Yeah probably leaving it, as this current simplification is non-recursive and doesn't actually simplify everything.

PR #250, opened 2023-08-04
Macro for pass pipeline specification

rachitnigam:
Adds a bunch of macros to clean up the pass specification process and replaces the `--show-ir` flag with `--dump-after` which allows us to control exactly which pass to print out information after. Additionally, the macro for pass specification tracks the amount of time each pass took at `--log info`.


gabizon103 (approved):
looks good!

PR #252, opened 2023-08-04
Fixes a small bug in expr interning

### LINKED ISSUE #251: [IR] Interval checking generates invalid constraint
For the following program:
```
comp Foo[#W]<G: 1>(
  @[G, G+1] in: #W,
) -> (
  @[G, G+1] out: #W,
) where #W > 0 {
  out = in;
}

comp TradeoffDiv[#W, #I, #K]<G: 1>(
  @interface[G] go: 1,
  @[G, G+1] left: #W,
) -> (
  @[G+#W, G+#W+1] out: #W,
) where
  #I > 0,
  #K > 0,
  #W == #I * #K
{
    i := new Foo[#W]<G>(left);
    bundle o_acc[#I+1]: for<#a> @[G+#a*#K, G+#a*#K+1] #W;
    o_acc{0} = i.out;
}
```

We get the error:
```
error: source port does not provide value for as long as destination requires
   ┌─ fail-bundle.fil:21:16
   │
21 │     o_acc{0} = i.out;
   │     --------   ^^^^^ source is available for @[G, G+1]
   │     │           
   │     requires value for @[G+#K, G+#K+1]
   │
   = Cannot prove constraint: #_2 >= 0 & 1 > #_2 & #a >= 0 & 1 > #a => #K >= 0 & 1 >= #K+1
   = Counterexample: #K = 1 (unmentioned parameters are 0)
```
The problem seems to be that the liveness of `o_acc{0}` is computed to be `[G+#K, G+#K+1]` but it should be `[G, G+1]`.

UnsignedByte:
Expr interning used to assume when one operator in a binary operation was `0` to return the other. This is only true for addition, and should be treated differently for different operators, which is fixed here. Fixes #251.


rachitnigam:
Was this the cause of #251?

test.fil
rachitnigam, line 1: Let’s not commit this filr

rachitnigam (approved):
LGTM!

PR #253, opened 2023-08-04
Removes some duplicated ir tests and unnecessary tests.

UnsignedByte:


rachitnigam:
Can you update the issue description and explain which tests are being removed and why?

rachitnigam:
Do we even want to keep the `tests/compile/ir` directory or just reuse the expect files from the old flow?

UnsignedByte:
This can be closed because the tests are already going to be changed by #256.

Issue #222, opened 2023-07-16
Change visitor interface to include context info.


UnsignedByte:
Now that `Foreign`s have been added, we should add `Context` to all the visitor functions which is necessary to resolve these ports.

rachitnigam:
@UnsignedByte is this addressed by the PR?

PR #254, opened 2023-08-04
Changes visitor pass and makes bundle elim implement it.

UnsignedByte:
Adds context info to visitor trait, also resolving #222. Also makes bundle-elim implement visitor after these changes.


src/ir_visitor/visitor.rs
```diff
@@ -217,27 +260,29 @@ where
         };
 
         // Traverse the commands
-        let mut cmds = std::mem::take(&mut comp.cmds);
-        match self.visit_cmds(&mut cmds, comp) {
+        let mut cmds = std::mem::take(&mut ctx.get_mut(idx).cmds);
```
rachitnigam, line 263: The way this is currently structured extracts the commands out from the component but gives you mutable access to the context. What I was thinking originally is that we instead completely "detach" the component from the context (by defining a `swap` method on `IndexStore`) and then providing mutable access to the component but immutable access to the context.

UnsignedByte: To simplify the number of arguments then could we re-add the "idx()" function on components? BundleElim requires component idx information which currently would have to be passed in separately from everything else.

rachitnigam: Oh, I see. Yeah, we can try that but that was causing annoying build problems because we’d need a default idx

rachitnigam: One other option is to abstract all of this using a `VisitData` struct which is also more future proof than adding more arguments. Calyx visitor has the same problem where we've slowed made the visitor interface more and more complicated with more arguments

UnsignedByte: Personally more of a fan of the second option! I'll work on that next.

src/ir_visitor/visitor.rs
```diff
@@ -29,16 +32,30 @@ impl Action {
     }
 }
 
+/// Contains information passed to visitor functions
+pub struct VisitorData<'comp> {
+    pub ctx: &'comp ir::Context,
+    pub comp: ir::Component,
+    pub idx: ir::CompIdx,
```
rachitnigam, line 39: Comments for these fields.

src/ir_visitor/visitor.rs
```diff
@@ -29,16 +32,30 @@ impl Action {
     }
 }
 
+/// Contains information passed to visitor functions
+pub struct VisitorData<'comp> {
```
rachitnigam, line 36: Is the lifetime name correct? I thought the lifetime is that of the entire pass and not just the component?

rachitnigam: Oh, I guess you have to change the context after traversing each component. Maybe say that in the comment?

rachitnigam (approved):
One comment about adding more docs! Merge after addressing and once #255 is done!

rachitnigam (approved):
LGTM!

PR #255, opened 2023-08-05
Extract max-states pass.

### LINKED ISSUE #227: Extract max_states calculation for `--dump-interface` flag.
> The max states computation needs to be moved out somewhere (maybe as a method on `Component`) because it is also used to generate `INTERFACE.json` when we're running our test harness. Currently everything works out because `fud/filament.py` doesn't use `--ir` with `--dump-interface` but it will break when we remove the ast-based compilation.
> 
> We can leave it here for now and extract in a following PR but please open an issue.

_Originally posted by @rachitnigam in https://github.com/cucapra/filament/pull/186#discussion_r1269806129_

UnsignedByte:
Fixes #227.

### DIFF
--- a/src/ir_passes/lower/utils.rs
+++ b/src/ir_passes/lower/utils.rs
@@ -1,8 +1,11 @@
+use std::collections::HashMap;
+
 use crate::ir::{
-    self, CompIdx, Component, Context, Ctx, EventIdx, ExprIdx, ParamIdx,
-    PortIdx,
+    self, CompIdx, Component, Context, Ctx, DenseIndexInfo, Event, EventIdx,
+    ExprIdx, ParamIdx, Port, PortIdx,
 };
 use calyx_ir as calyx;
+use linked_hash_map::LinkedHashMap;
 
 type AttrPair = (calyx::Attribute, u64);
 /// A set of interface ports that are required for all components.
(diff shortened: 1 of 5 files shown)


rachitnigam:
To make sure this actually works, we should add an option to the `fud` execution to use the IR-based max states as well: https://github.com/cucapra/filament/blob/main/fud/filament.py#L181

rachitnigam:
Can we instead merge this into #256 and replace the `max-states` in the old flow with this? That PR doesn't have any of the old flow at all.

src/ir_passes/lower/utils.rs
```diff
@@ -78,11 +79,39 @@ pub(super) fn port_name(
...
 pub(super) fn comp_name(idx: CompIdx, ctx: &impl Ctx<Component>) -> String {
     ctx.get(idx)
         .src_info
         .as_ref()
         .map(|src| src.name.to_string())
         .unwrap_or_else(|| format!("comp{}", idx.get()))
 }
+
+/// Calculates the max states used for every fsm.
+pub fn max_states(idx: CompIdx, ctx: &Context) -> LinkedHashMap<EventIdx, u64> {
```
rachitnigam, line 92: Why is the key `EventIdx` and not `Foreign<EventIdx>` or something? Since this computation happens over every component, we're going to end up overwriting the wrong `EventIdx`

rachitnigam: Oh, my bad, I guess it's computing it for all the events in a specific component. The `ctx` parameter confused me.

UnsignedByte: This function only calculates max states for one component, becuase dump interface only requires it to be calculated for main, and its calculated on demand in lower as well.

rachitnigam: Can we just pass the `ir::Component` by reference to this method instead of passing in the context?

src/ir_passes/mono/monosig.rs
```diff
@@ -395,43 +395,38 @@ impl MonoSig {
...
+                 events,
+             }| {
+                ir::InterfaceSrc {
+                    name,
+                    ports,
+                    interface_ports: interface_ports
+                        .into_iter()
+                        .map(|(ev, id)| (*self.event_map.get(&ev).unwrap(), id))
+                        .collect(),
+                    params: if underlying.is_ext {
```
rachitnigam, line 413: Move this `if` out into a local variable

rachitnigam (approved):
Minor comments on style. Merge at your discretion after addressing them!

Issue #189, opened 2023-06-24
Change availability interval syntax


rachitnigam:
Currently, we have an inconsistent syntax for availability intervals in the language:
```
comp Foo(@[G, G+1] l: 32) {
  bundle f[10]: for<#i> [G, G+#i] 32;
}
```

I think we should switch to the following syntax for ports to make things more consistent:
```
comp Foo(l: [G, G+1] 32) { 
  bundle f[10]: for<#i> [G, G+#i] 32;
}
```

This has a couple of benefits:
1. Gets rid of the syntactic noise of `@`
2. Once we have type inference, we can elide the `[G, G+1]` and have syntax that looks pretty much like a normal language.

PR #257, opened 2023-08-05
`let`-bound parameters

rachitnigam:
Supersedes #209

Monomorphization is the tricky bit and will require some way to mark the end of a scope so that bindings from a particular parameter is deleted and can be updated if need (for example for the next iteration of a loop).


rachitnigam:
@UnsignedByte feel free to take over this PR if you're interesting in getting this feature landed soon

src/ast/component.rs
```diff
@@ -42,6 +42,7 @@ impl Component {
                 | Command::Bundle(_)
                 | Command::If(_)
                 | Command::ForLoop(_)
+                | Command::PortLet(_)
```
rachitnigam, line 45: Port let is not needed because it is the same as a bundle of size one. #189 implements the desugaring needed to support let bound ports directly using the existing bundle infrastructure.

src/ir_passes/mono/monodeferred.rs
```diff
@@ -266,6 +266,26 @@ impl<'a, 'pass: 'a> MonoDeferred<'a, 'pass> {
...
+        // Inserts this param into the binding.
+        self.monosig.binding.insert(
+            Underlying::new(param),
+            self.monosig
+                .base
+                .bin(self.monosig.base.get(expr.idx()).clone())
+                .as_concrete(&self.monosig.base)
+                .unwrap(),
+        );
+    }
```
rachitnigam, line 287: Not sure if this works when you have a local parameter dependent on a loop index.

rachitnigam:
@UnsignedByte is this ready to review/merge? The conflicts are probably from the recent merge of #266

UnsignedByte:
> @UnsignedByte is this ready to review/merge? The conflicts are probably from the recent merge of #266

Yes, it should be ready, I couldn't request a review I think because you originally created the PR?

I'll fix the conflicts when I can probably tomorrow morning though.

rachitnigam:
Okay, thanks! I've fixed the conflicts.

src/ir_passes/mono/monodeferred.rs
```diff
@@ -227,11 +230,17 @@ impl<'a, 'pass: 'a> MonoDeferred<'a, 'pass> {
...
             let index = Underlying::new(*index);
             self.monosig.binding.insert(index, i);
+            let mut nlets: u32 = 0; // count p_lets as they generate assignments we have to pop later.
             for cmd in body.iter() {
+                nlets += matches!(&cmd, ir::Command::Let(_)) as u32;
                 let cmd = self.command(cmd);
                 self.monosig.base.cmds.extend(cmd);
             }
-            self.monosig.binding.pop();
+            for _ in 0..nlets {
```
rachitnigam, line 239: This should probably use [`truncate`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.truncate) or something instead of using a loop to repeatedly pop.

rachitnigam (commented):
LGTM! One thing that occurs to me is that we probably don't implementing shadowing correctly now that we have `let`-bound variables. This is, in general, an annoying problem to solve and will require some redesign of `astconv`. We might want to take ideas from the Rust compiler's approach to [scope representation using the "ribs" abstraction](https://rustc-dev-guide.rust-lang.org/name-resolution.html?highlight=scope#scopes-and-ribs).

PR #258, opened 2023-08-07
Mono fixes

gabizon103:
Adds the types `Base<T>` and `Underlying<T>` for use in monomorphization, which are wrappers around an `Idx<T>`. A lot of functions that had `Idx` types as inputs and outputs now have `Base` or `Underlying`. 

This does not change any of the functionality of the code, the only purpose is to make the code easier to understand. Previously, looking at comments or carefully reading the code were the only ways to tell what each `Idx` belonged to, but now with these wrapper types it is in the code itself, in a way that makes it obvious. Ideally, this will make it easier if we ever have to revisit this for debugging or expanding features. The code is a bit more verbose at times, but I don't think there is any way around this if we want to make the distinction obvious.


rachitnigam:
Once #255 is merged, the tests for this should pass!

src/ir_passes/mono/utils.rs
```diff
@@ -0,0 +1,70 @@
+use crate::utils::Idx;
+
+/// Wraps an Idx that is meaningful in the base component
```
rachitnigam, line 3: Awesome!! Can we add a little more documentation for these structs to describe what exactly base and underlying components mean and how they show up in the passes?

rachitnigam (approved):
I'm a fan of this change and I think we should merge it! One wandering remark here is that it's kind of a shame that we have to explicitly call the `.idx()` function when call the `.get` method on the components. I was wondering if there was some way to implement a trait that would do the unwrapping for us and make it even more impossible to pass the indices into the wrong places. The high-level idea would be:

1. The `underlying` component is wrapped in `Underlying` or `UnderlyingComp` and the `Underlying<T>` indices would only work on that.
2. The `base` component would get to use the same `Ctx` method it always uses.

For example:
```rust
struct UnderlyingComp<'a>(&'a Component);

trait UnderlyingCtx {
  fn get(&self, k: Underlying<T>) -> &T
  // We don't mutate Underlying so no need for add?
}

// This is the magical `impl`. It says that for every type that component is a context for, we also want it to be an UnderlyingCtx
impl<'a, T> UnderlyingCtx<T> for UnderlyingComp<'a>
  where ir::Component: Ctx<T> {
  fn get(...) {
     /// The .idx() method is only visible in this file
    (self.0 as Ctx<T>).get(k.idx())
  }
}
```

The net effect of this trick can be that the `.idx()` method on `Underlying` is no longer public and the only way to use `Underlying<T>` is using the `.get()` method defined by `UnderlyingCtx`.

This is a big change so I'd recommend merging what you have here before thinking about this other extension.

gabizon103:
Hmm, yeah I was actually thinking that also, but wasn't sure how much we wanted these wrappers to "interact" with the existing IR. But I agree that it would be nice to use `get` without doing the unwrapping ourselves, so I'll try implementing something like that after I merge this. Thanks!

PR #260, opened 2023-08-09
IR phantom check

gabizon103:
Implements the phantom check pass in the IR, adapted from the old AST flow. The structure is pretty much identical to the old pass. 

All the tests that were passing previously still pass, and I added two more tests that were present in the old flow to check for the two ways phantom check could fail. The error messages produced are identical to the old ones.


src/ir/comp.rs
```diff
@@ -22,8 +22,8 @@ impl Context {
     }
 
     /// Add a new component to the context
-    pub fn comp(&mut self, is_ext: bool) -> CompIdx {
-        let comp = Component::new(is_ext);
+    pub fn comp(&mut self, is_ext: bool, is_entry: bool) -> CompIdx {
```
rachitnigam, line 25: Why is this needed? The existence of `is_entry` indicates that the component is an externally visible one.

rachitnigam:
@gabizon103 just checking if this should be reviewed or not

rachitnigam (commented):
Overall looks good but I'm going to loop in @UnsignedByte to make sure the `InterfaceSrc` change makes sense. From what I understand, previously, the interface port information was only added for externally visible components but now its added for all components?

As I write this, I'm realizing that can't be true because lowering needs to have access to the interfaces for all components that are being compiled so where does lowering get that information from (the mapping from an event to its corresponding interface port).

rachitnigam:
Looks like `Event` defines the `has_interface` field to track whether there is an interface port for a component? Is the extra information useful for generation error message?

gabizon103:
Ah I missed that, that should probably be enough - should I just use that instead?

rachitnigam:
Yeah

src/ir/comp.rs
```diff
@@ -233,6 +233,12 @@ impl Component {
 
         sig_params
     }
+
+    pub fn phantom_events(&self) -> impl Iterator<Item = EventIdx> + '_ {
```
rachitnigam, line 237: Documentation comments needed here

src/ir/control.rs
```diff
@@ -104,6 +104,12 @@ impl InvIdx {
         inv.inst
     }
 
+    /// The times the invoke uses, along with the EventBind infos
+    pub fn times(self, ctx: &impl Ctx<Invoke>) -> Vec<(TimeIdx, InfoIdx)> {
```
rachitnigam, line 108: This is a little funky function to expose in the interface. Are you expecting it to be used in other places? Otherwise we can probably do:
```
comp.get(inv).events...
```

If we decide to keep it, pro-tip about interface design, instead of `.collect` in this method, you can actually return the iteration directly using:
```
times(...) -> impl Iterator<(TimeIdx, InfoIdx)> 
```

This is useful because often times, when you get a `Vec` out, you immediately want to iterate using it anyways.

src/ir/from_ast/astconv.rs
```diff
@@ -375,7 +375,6 @@ impl<'prog> BuildCtx<'prog> {
 
         // If this is a signature port, try adding it to the component's external interface
         if is_sig_port {
-            // If the component is expecting interface information, add it.
```
rachitnigam, line 378: Let's keep this comment?

src/ir_passes/lower/utils.rs
```diff
@@ -80,7 +81,7 @@ pub(super) fn port_name(
 }
 
 /// Returns the name of an [Component]
-pub(super) fn comp_name(idx: CompIdx, ctx: &impl Ctx<Component>) -> String {
+pub(super) fn comp_name(idx: CompIdx, ctx: &ir::Context) -> String {
```
rachitnigam, line 84: Any reason for this change?

src/ir_passes/phantom_check.rs
```diff
@@ -0,0 +1,127 @@
...
+                        .inst_loc;
+                    let err =
+                        Error::malformed("reuses instance uses phantom event for scheduling")
+                    .add_note(self.diag.add_info("invocation uses phantom event", bind_info))
+                    .add_note(self.diag.add_info("event is a phantom event", event_info))
+                    .add_note(self.diag.add_info("previous use", prev_inv_info))
+                    .add_note(self.diag.add_info("phantom ports are compiled away and cannot be used for resource sharing", GPosIdx::UNKNOWN));
+                    self.diag.add_error(err);
+                    self.diag_count += 1;
+                    return Action::Stop;
```
rachitnigam, line 71: If we return `Stop` here, we won't be able to report other phantom check error message in this component right?

src/ir_passes/phantom_check.rs
```diff
@@ -0,0 +1,127 @@
...
+/// Phantom events are valid iff:
+/// 1. The component doesn't share any instances
+/// 2. The component doesn't use an subcomponents that need to use the
+///    corresponding event in their interface, i.e., the uses of the event are all phantom
+pub struct PhantomCheck {
+    instance_used: HashSet<ir::InstIdx>,
+    prev_invoke: Option<ir::InvIdx>,
+    phantom_events: Vec<ir::EventIdx>,
+    diag: diagnostics::Diagnostics,
+    diag_count: u32,
```
rachitnigam, line 23: The `Diagnostics` struct returns a count of the errors when you call `report_all` on it. The general idea with it is you build up all the errors in a pass and then report them together. By default, if a pass implements `Default`, we [create a new instance of the struct](https://www.irisa.fr/alf/downloads/ADA/p127-josipovic.pdf) every time we visit a new component so the `diag` object is not being shared here. To share information, you have to manually override `clear_data`.

We probably need to do some restructuring to make sure we report *all* phantom check problems in all components using this pass. Maybe a test where we have two components with problems and at least one of them has two phantom check violations would be useful.

rachitnigam:
@gabizon103 I ended restructuring the code a bit more to allow for multiple errors to be captured. Also, I realized that the original compiler flow was missing an important case for phantom check: if an invocation is used in a loop but its corresponding instance is not defined in the same loop, after unrolling, we'll have a reuse problem. 

Take a look at the code and see if it makes sense. I'm going to merge things soon but let me know if you have any comments/

Issue #190, opened 2023-06-24
Bindings for signatures


rachitnigam:
Filament type signatures can get quite messy:
```Rust
comp GenDeserialize[#W, #N, #B, #C, #H]<G: #C*#N+#H-2>(
    in[#N*#B]: for<#p> @[G+#C*(#p/#B), G+#C*(#p/#B)+#H] #W
) -> (
    out[#N*#B]: for<#p> @[G+#C*(#N-1), G+#C*(#N-1)+#H] #W
) where #N > 1, #B > 0, #C > 0, #H > 0
```

We should use #188 to make it easier to read. The idea is that we allow the user to `let`-bind things for the scope of the signature:
```Rust
comp GenDeserialize[#W, #N, #B, #C, #H]<G: #Del>(
    in[#Tot]: for<#p> @[G+#C*(#p/#B), G+#C*(#p/#B)+#H] #W
) -> (
    out[#Tot]: for<#p> @[G+#Max), G+#Max+#H] #W
) with {
  let #Tot = #N*#B;
  let #Max = #C*(#N-1);
  let #Del = #C*#N+#H-2;
} where #N > 1, #B > 0, #C > 0, #H > 0
```

The `with` construct defines the parameters in the scope of *both* the signature and the body of the component. This doesn't fix the other problem with parameters in Filament which is the syntactic noise of `#`. 

Maybe we should just let you write variables without `#` and disambiguate. The problem with this approach originally is that we can't distinguish between events and parameters which means the frontend cannot create `ast::Time` structs without some type disambiguation.

PR #262, opened 2023-08-10
Sig bindings

gabizon103:
#190


rachitnigam:
@gabizon103 any luck debugging the failing test?

gabizon103:
@rachitnigam Haven't had much of a chance yet, but that's what I'm going to do today!

rachitnigam:
@gabizon103 it is occurring to me that a possible problem can be that we're not defining the parameters in the `with` section use a subcomponent:
```
comp foo[#W](in: [G, G+1] #B) -> () with {
  #B = #W + 2
} { ... }
comp bar(...) {
  f := new foo[10]; // This should define both #W and #B
}
```

If the `#B` parameter is not defined within `bar`, then the port `f.in` is going to use an unknown parameter causing problems. In general, all parameters in the `with` section must be visible to the parent component.

gabizon103:
> If the #B parameter is not defined within bar, then the port f.in is going to use an unknown parameter causing problems. In general, all parameters in the with section must be visible to the parent component.

Yeah I've come to this conclusion also. I am trying to figure out how to do this - `astconv` is pretty dense and I haven't looked at it much before this

rachitnigam:
I've mucked up the PR because of this merge. If you're not working on it in the next hour or two, I can take a shot and you can try getting #263 PR merged in the meantime?

gabizon103:
Sure, I'll work on #263!

rachitnigam:
Subsumed by #294.

PR #263, opened 2023-08-11
Base/Underlying Ctx

### LINKED ISSUE #259: Make mono wrapper usage more strict
There are now wrappers around underlying and base `Idx`s in monomorphization (#258), but we have to unwrap them before we look them up anywhere, so it is still possible to look up an `Idx` in the wrong component. It would be cool to have it such that you don't need to do the unwrapping, but instead can just look up some `Underlying` in an `UnderlyingCtx`, and looking up a `Base` in an `UnderlyingCtx` would error. Change outlined [here](https://github.com/cucapra/filament/pull/258#pullrequestreview-1566241378).

gabizon103:
Closes #259 
Implements `BaseCtx` for `BaseComp` and `UnderlyingCtx` for `UnderlyingComp`, so that we don't have to unwrap `Idx`s before looking them up, and looking up an `Idx` in a context where it doesn't belong will error. I think in general this makes it easier to follow what's happening and harder to mess up when writing code, but I guess I'll put that to test when implementing some of the language extensions. 

We do have to do some unwrapping ourselves when we want to create new structures that aren't stored in a component's `IndexStore` fields, e.g. `Range`s. We will monomorphize its `start` and `end` fields, and then get the corresponding `Base<Time>`s, then unwrap those when creating the new `Range`. I'm not sure that there is a good way around this, and I think this is fine. 

In some cases we want to modify some existing parts of the `BaseComp` (like add a `Command` to the existing `Command`s), so I exposed some functions for doing this instead of making the component itself visible, which seems like a bad idea.  

~~The part where things get a little strange is when we have to interact with things that are not mono-specific, such as the old `Context` that we are monomorphizing. When we call `Monomorphize::next()`, we get a `(BaseComp, Base<Component>)`, where `BaseComp` is newly-monomorphized component and `Base<Component>` is a wrapper around the component `Idx` that is meaningful in the new context. At this point `Base<Component>` points to a default (empty) component, so the idea is that we look it up in the new context, get the aforementioned default `Component`, and then call `mem::swap` on the two. The reason this got a little weird was because we were swapping `Component`s and not `BaseComp`s, and there was no way to rip the component out of its `BaseComp` wrapper - you can get a reference to it (for the purpose of passing it into methods such as `ExprIdx::as_concrete()`), but that is it. What I ended up doing was implementing `Clone` for components so we could clone the thing inside of the `BaseComp` and then swap that in, but this seems expensive/wasteful. Is there a better way to do this?~~

To solve the issue above, I implemented a `swap` function inside `BaseComp` that can peek inside the `BaseComp` that is passed in and do the swap there. This eliminates the need for cloning the component, and then swapping.


src/ir_passes/mono/utils.rs
```diff
@@ -1,4 +1,125 @@
...
+    }
+    pub fn func(&mut self, expr: ir::Expr) -> Base<ir::Expr> {
+        Base::new(self.0.func(expr))
+    }
+    pub fn resolve_prop(&mut self, prop: ir::Prop) -> Base<ir::Prop> {
+        Base::new(self.0.resolve_prop(prop))
+    }
+}
+
+pub trait BaseCtx<T> {
```
rachitnigam, line 93: Can we change the `Ctx` and `MutCtx` traits to take an `IdxLike` instead of operating on `Idx<T>`? This will allow us to implement `Ctx` for these new structs because `Base` and `Underlying` already implement `IdxLike`. The trick to doing this is changing the `Ctx` trait like this:
```rust
pub trait Ctx<T, K=Idx<T>> where K: IdxLike<T> {
    /// Add a new value to the context
    fn add(&mut self, val: T) -> K;
    /// Get the information associated with a value
    fn get(&self, idx: K -> &T;
}
```

This will allow all of the existing `Ctx` implementation to continue working because they will use the default value of `K`. However, for `BaseComp`, we can say:
```
impl Ctx<T, Base<T>> for BaseComp
  where ir::Component: Ctx<T> {
  fn get(&self, k: Base<T>) -> &T { self.0.get(k.get()) }
```

This is definitely getting to the galaxy brain meme levels of type magic so feel free to say no to this.

rachitnigam:
Overall, the design looks good. I left one comment about the new `Ctx` implementations.

> The reason this got a little weird was because we were swapping Components and not BaseComps, and there was no way to rip the component out of its BaseComp wrapper - you can get a reference to it (for the purpose of passing it into methods such as ExprIdx::as_concrete()), but that is it.

The problem here seems to be that there is in fact a `BaseContext` and `UnderlyingContext` too but it might be tedious to represent them as such. It is a little tricky, having to clone the components, but one way around it is having a function defined in the same file as `BaseComp` that gets to look into the definition of `BaseComp` and efficiently implement the `swap` for you. The requirement would be that it takes a `self` and returns a value of type `Self`. Could that work?

gabizon103:
Yeah, I was thinking about how far I wanted to go in terms of implementing an actual `BaseContext` and `UnderlyingContext`, and it seemed tedious to do it. I think the the `swap` approach you mentioned is worth exploring, so I'll try doing that.

src/ir_passes/mono/utils.rs
gabizon103: Lol maybe, but I'll try it out!

gabizon103:
@rachitnigam I implemented the change for `Ctx` and also implemented `swap` inside `BaseComp` so we don't have to clone it

src/ir/ctx.rs
```diff
@@ -1,21 +1,27 @@
 use super::Component;
-use crate::utils::Idx;
+use crate::utils::{Idx, IdxLike};
 
 /// A context for storing values with their indices.
 /// The context is indexed by [`Idx<T>`].
-pub trait Ctx<T> {
+pub trait Ctx<T, K = Idx<T>>
+where
+    K: IdxLike<T>,
```
UnsignedByte, line 8: The `IdxLike` stuff is also implemented in #288 in pretty much the exact same way so merging this will likely cause conflicts with that (no changes needed, just a note).

src/ir_passes/mono/monomorphize.rs
```diff
@@ -171,11 +171,13 @@ impl Monomorphize<'_> {
...
 
         // Build a new context
-        while let Some((comp, idx)) = mono.next() {
+        while let Some((mut comp, idx)) = mono.next() {
             let default = mono.ctx.get_mut(idx.get());
-            let mut comp = comp.comp().clone();
-            std::mem::swap(&mut comp, default);
-            let val = ir::Validate::new(&comp, &mono.ctx.comps);
+            // let mut comp = comp.comp().clone();
+            // std::mem::swap(&mut comp, default);
```
UnsignedByte, line 177: Delete comments

UnsignedByte (approved):
LGTM!

src/ir_passes/mono/utils.rs
UnsignedByte: Also done in #288

src/ir/ctx.rs
gabizon103: Thanks!

rachitnigam:
Let's merge #288 first and then this? Sorry for the conflicts created @gabizon103

PR #271, opened 2023-08-13
Activation literal based SMT checking

rachitnigam:
Implements activation literal based SMT checking. The big change is using a bool to create an implication instead of using `push` and `pop`. This allows the solver to reuse work it did to solve previous queries.

Currently using my `easy-smt` fork. Once https://github.com/elliottt/easy-smt/pull/21 is merged, we can switch to the original package.


rachitnigam:
Okay, the act lit nonsense didn't work at all so instead this PR attempts to discharge all the assertions in a component together. SMT solvers are optimized for giant queries anyways so this might work better.

rachitnigam:
Okay, finally this works! @UnsignedByte & @gabizon103 this adds an option to configure the solver we’re using and switches the default to cvc5. We can add a flag to runt to configure which silver is being locally. If you still can’t run this program locally, I’ve added a flag to skip SMT checks entirely.

PR #283, opened 2023-08-24
Fixes fsms

### LINKED ISSUE #282: Bug in slow fsm generation
Slow fsm generation currently fails with the iterative-divider test (mentioned in https://github.com/cucapra/filament/pull/279#issuecomment-1685303108).

UnsignedByte:
Fixes #282.

Fixed an issue where the order of non-exclusive guards led to incorrect behavior.

Counter fsms were incremented when `go || state != 0` and reset to zero when `state == max_state` but the resultant guard was generated `go || state != 0 ? add_1 : state == max_state ? 0 : x` which led to the counters counting up indefinitely.


tests/compile/fsm-guard.fil
```diff
@@ -0,0 +1,54 @@
...
+    in: [G, G+#DELAY] 32,
+) -> (
+    out[#DELAY]: for<#i> [G+#i, G+#i+1] 32
+) where #DELAY > 0 {
+    for #ii in 0..#DELAY {
+        out{#ii} = in;
+    }
+}
+
+comp test[#DELAY, #NUM]<G: #DELAY>(
```
rachitnigam, line 13: Any chance we could give this component a slightly more informative name? If not, maybe a comment explaining what it is doing within the test?

tests/compile/fsm-guard.fil
```diff
@@ -0,0 +1,54 @@
+// Tests the guards of an fsm.
+
+comp test2[#DELAY]<G: #DELAY>(
```
rachitnigam, line 3: ALL CAPS parameters look pretty ugly but we haven't really discussed any good conventions around naming schemes. Maybe we should do that in a new doc page to ensure consistency and readable code for everyone.

rachitnigam (approved):
LGTM! Could you explain what the problem ended up being? I think it would be useful in the since slow FSMs are a little tricky.

PR #285, opened 2023-08-28
Enable preservation of compiled names.

UnsignedByte:
Adds a flag to use infos to generate names during compilation.

### DIFF
--- a/src/ir_passes/lower/utils.rs
+++ b/src/ir_passes/lower/utils.rs
@@ -23,25 +23,29 @@ pub(super) const INTERFACE_PORTS: [(AttrPair, (&str, u64, calyx::Direction));
 pub(super) fn interface_name(
     idx: EventIdx,
     comp: &Component,
+    debug: bool,
 ) -> Option<String> {
-    if !comp.get(idx).has_interface {
-        return None;
-    }
+    let ev = comp.get(idx);
 
-    Some(
+    ev.has_interface.then(|| {
         comp.src_info
             .as_ref()
             .map(|src| src.interface_ports.get(&idx).unwrap().to_string())
-            .unwrap_or_else(|| format!("ev{}", idx.get())),
-    )
+            .or_else(|| debug.then(|| ev.info.get_name(comp)).flatten())
+            .unwrap_or_else(|| format!("ev{}", idx.get()))
+    })
 }
 
 /// Converts an [ir::ExprIdx] into a [calyx::Width].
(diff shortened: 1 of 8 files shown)


UnsignedByte:
This might break if infos are cloned, so I will probably use idxs to ensure uniqueness as well.

UnsignedByte:
We should probably find a way to test this, but I'm not sure exactly how.

rachitnigam:
Can we just add a pass test that generates the Calyx IR output and commit the generated file as the golden output?

UnsignedByte:
> Can we just add a pass test that generates the Calyx IR output and commit the generated file as the golden output?

Yes but this file would change if any changes in the passes changes IR or behavior in any way (such as through optimizations or even just any datapath changes that affects the exact state of the IR before compilation)

UnsignedByte:
> Yeah, c'est la vie. The other option is defining a Calyx-like pass manager so we can run a specific pass and commit the output from the pretty printer.

Alright, I'll do the former for now as it is a lot simpler.

UnsignedByte:
Quickly implemented this but there's already a big problem: calyx doesn't guarantee attribute order to be constant, so the resultant compiled code is not guaranteed to be the same each time. There are likely also some `HashMap` issues that have to be fixed but the main problem is that calyx itself cannot guarantee consistent behavior.
image

src/ir_passes/lower/utils.rs
```diff
@@ -71,14 +85,27 @@ pub(super) fn port_name(
...
+            base.apply(|p, c| port_name(p, ctx, c, debug), ctx)
         }
-        ir::PortOwner::Local => format!("p{}", idx.get()),
+        ir::PortOwner::Local => debug
+            .then(|| info_name(comp.get(idx).info, comp))
+            .unwrap_or_else(|| format!("p{}", idx.get())),
     }
 }
 
+pub(super) fn instance_name<C: Ctx<Instance> + Ctx<Info>>(
```
rachitnigam, line 99: Could we define these methods on some relevant struct so that we don't have to thread through the `debug` argument?

rachitnigam (approved):
LGTM! Minor comment that you can resolve at your discretion and merge.

rachitnigam:
Oh, also, I should mention: I think Calyx does guarantee the right attribute order insofar as Calyx itself uses these kinds of golden tests. I think we might be using a HashMap somewhere?

PR #288, opened 2023-08-29
Implement Ctx for Foreigns.

### LINKED ISSUE #261: Implement `Ctx` for `Context` for resolving `Foreign`
The `ir::Context` structure is the only thing that can resolve `Foreign` indices since we need both the component and the underlying index type. Providing convenience `get` methods for all the various `Foreign` in the IR would be useful. However, the current `Ctx` trait is not amenable to this change since we need to implement `add` as well. It might be useful to separate the trait out so that we can implement `get` methods.

UnsignedByte:
Fixes #261 and also generalizes some existing functions to use `Ctx` instead of `Component`

### DIFF
--- a/src/ir/comp.rs
+++ b/src/ir/comp.rs
@@ -1,7 +1,7 @@
 use super::{
-    Cmp, CmpOp, Command, Ctx, Event, EventIdx, Expr, ExprIdx, Fact, IndexStore,
-    Info, InfoIdx, InstIdx, Instance, Interned, InvIdx, Invoke, MutCtx, Param,
-    ParamIdx, Port, PortIdx, Prop, PropIdx, Time, TimeSub,
+    AddCtx, Cmp, CmpOp, Command, Ctx, Event, EventIdx, Expr, ExprIdx, Fact,
+    IndexStore, Info, InfoIdx, InstIdx, Instance, Interned, InvIdx, Invoke,
+    MutCtx, Param, ParamIdx, Port, PortIdx, Prop, PropIdx, Time, TimeSub,
 };
 use crate::{ast, utils::Idx};
 use fil_derive::Ctx;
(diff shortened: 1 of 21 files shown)


src/ir/utils.rs
```diff
@@ -578,6 +582,36 @@ where
...
+
+    fn new(idx: usize) -> Self {
+        Self {
+            key: Idx::new(idx),
+            owner: Idx::UNKNOWN,
+        }
+    }
+
+    fn get(self) -> usize {
+        self.key.get()
```
UnsignedByte, line 602: This should be discussed as it seems somewhat dangerous, but I can't think of a better method that is useful.

src/ir/comp.rs
```diff
@@ -33,36 +33,44 @@ impl InterfaceSrc {
...
 
     // Component defined values.
     #[ctx(Port)]
+    #[add_ctx(Port)]
     #[mut_ctx(Port)]
     /// Ports and bundles defined by the component.
     ports: IndexStore<Port>,
     #[ctx(Param)]
+    #[add_ctx(Param)]
     #[mut_ctx(Param)]
```
UnsignedByte, line 57: It also might be worth combining this somehow, maybe with a macro for all ctx types or something?

src/ir/ctx.rs
```diff
@@ -1,21 +1,34 @@
 use super::Component;
-use crate::utils::Idx;
+use crate::utils::{Idx, IdxLike};
 
 /// A context for storing values with their indices.
-/// The context is indexed by [`Idx<T>`].
-pub trait Ctx<T> {
+/// The context is indexed by [`IdxLike<T>`].
+pub trait AddCtx<T, I = Idx<T>>
```
gabizon103, line 6: Oh cool, I like separating this out. Was thinking of doing this with the mono wrappers so I'll use this once it's merged

gabizon103 (approved):
This looks good to me!

src/ir/comp.rs
rachitnigam: We can change the derive macro to instead support the syntax:
```
#[ctx(Param: add, get, mut)]
```
To signify that we want the `add`, `get`, and `mut` traits implemented.

src/ir/utils.rs
rachitnigam: `IdxLike` requires you to implement this method so we don't really have a choice but I see your point. Ideally we'd want to scope the methods of the trait so that the only way to use them is within the containers but trait methods are visible everywhere AFAIK.

rachitnigam (approved):
LGTM! Minor comments about possible future improvements.

Issue #192, opened 2023-06-26
Syntax change for parameter definitions and events.


UnsignedByte:
Parameters are currently much more common in expressions than events, so to clean up the syntax of filament we could remove the requirement for pound symbol in the `#PARAM` syntax and, to retain visual difference between events and parameters, add a prefix (`@` for example) to the event names in the syntax.

UnsignedByte:
This creates an ambiguity issue with `let` bindings: `let PARAM = PARAM2` now has no syntactic difference from `let port = port2`. Should we change `let port = port2` to `let port p= p2` or something like that? @rachitnigam

rachitnigam:
No, this is why it’s a non trivial change because it requires syntax disambiguating to figure out what is a port, a param, and an event.

rachitnigam:
Let me expand on the comment above a little bit. This *ambiguity* problem shows up almost every single language. For example, in Rust, type identifier, struct names, and variables all use the same syntax:
```rust
struct Foo;
type Bar = Foo;
fn baz(x: Bar) {
  let Bar = x;
}
```

The problem for the rust frontend is to look at this code and figure out the correct binding for each symbol. In this case, the *position* of each identifier exposes the binding chain we need to look at. When you look at `type <ID> = <Expr>`, the `Expr` can only mention other identifiers in the type name bindings. Similarly, when looking at `let <ID> = <Expr>`, we know that `Expr` comes from the identifiers defined as variables.

The reason Rust doesn't have to do what Filament does, which is have distinct looking identifiers for different kinds of terms, is because it separates the frontend from the core representation(s). Now that we also have that separation in Filament, what we should actually do is remove the syntactic difference between ports, params, and events completely and use `astconv` to resolve these things for us.

rachitnigam:
We've made this change!

PR #290, opened 2023-08-29
change event and param syntax

UnsignedByte:
Implements #192.


UnsignedByte:
https://github.com/cucapra/filament/issues/192#issuecomment-1698083759 mentions an issue with let-binding syntax that might need to be resolved.

rachitnigam:
See https://github.com/cucapra/filament/issues/192#issuecomment-1698466139 for the answer. This change will require more work and we should probably not differentiate between the syntax of events and parameters at all.

gabizon103 (approved):
This looks good to me, but in regards to the discussion in #192, I think it might be worth having a clear (and not too ugly) syntactic distinction between params and events? I kind of like keeping `@` to denote events and leaving parameters with no prefix, instead of leaving them all with no prefix. I don't feel super strongly about this so if you want to remove all the prefixes then that would be cool too. This was just a thought.

UnsignedByte:
Something like `@prod.G` might look better? I do think that having the `@` definitely helps as far as understanding a complex expression with parameters and events, as we can sort of disregard the event parts to an extent and only look at the offset.

rachitnigam:
Hm, but then it's not very clear which thing is the event: `prod` or `G`. We could steal the Rust syntax for associated types:
```
m0 := M<G>(in);
m0::Ready // <- refers to the Ready event associated with the m0 invoke
```

However, we'd still need some syntax to access existentially defined parameters.

rachitnigam:
Haven't reviewed yet but wanted to note that #304 might create conflicts.

src/ast/control.rs
```diff
@@ -325,7 +325,7 @@ impl From<If> for Command {
 #[derive(Clone)]
 /// The type of the bundle:
 /// ```
-/// for<#i> @[G+#i, G+#i+1] #W
+/// for<#i> '[G+#i, G+#i+1] #W
```
rachitnigam, line 328: nit: the `'` should be inside the interval

src/ast/control.rs
```diff
@@ -401,7 +401,7 @@ impl BundleType {
 #[derive(Clone)]
 /// Represents a bundle of wires with timing guarantees
 /// ```
-/// bundle f[10]: for<#i> @[G+#i, G+#i+1] #W;
+/// bundle f[10]: for<#i> '[G+#i, G+#i+1] #W;
```
rachitnigam, line 404: nit: same as above

rachitnigam (approved):
LGMT but we should make that the error messages generated in `ir::info` also use the new syntax. They manually construct strings for formatting parameters in some places so will look wrong.

PR #294, opened 2023-08-31
Sig bindings attempt 2

rachitnigam:
Trying #262 again because merge borked it.


rachitnigam:
Next bug seems to be when trying to use it in `TradeOffDiv` implementation. @gabizon103 do you want to take over from here? My goal is to use this feature in the standard library so we can be certain that it's implemented correctly.

rachitnigam:
The problem seems to be that `astconv` is not completely eliminating the sig-bound parameters and they show up monomorphization causing problems. The invariant to be maintained is that all parameters bound in `with` are eliminated by the time we convert into the IR. Here is an MRE:
```
comp Foo[#K]<G: 1>() -> (
  out: [G, G+1] #I,
) with {
   let #I = #K + 1;
} {
  for #p in 0..#I { }
}

comp main<G:10>() -> () {
    f := new Foo[10]<G>();
}
```

gabizon103:
Right now the IR for `tests/compile/sig-bind-port.fil` looks like this before monomorphization:
```
comp %comp0[#N]<G: 1>(
  in: for<#_: 1> [G, G+1] #N*2
) -> (
) {
}
comp main[]<G: 1>(
  in: for<#_: 1> [G, G+1] 10
) -> (
) {
  F = instance %comp0[5, 10];
  f0, f0.in = invoke F<G>;
  f0.in[0] = in[0];
}
```
so it looks like the sig-bound parameter is added to the sig-owned parameters of the component during `astconv`. This causes issues with the bindings we keep track of in monomorphization, so will try to fix in `astconv`

rachitnigam:
Hm, look at the input program. Looks like we're adding extra bindings to the instance of `%comp0`. I think that might be because we're building an extended binding in `astconv`.

rachitnigam:
Okay, looks like that was the problem! Super weird that an extra parameter was causing a port map to be monomorphized incorrectly. I tried to run the validator on the IR but that seems to cause a stack overflow. We should really fix the validate pass and create an option to run it after every single pass...

rachitnigam:
@gabizon103 looks like this is working now! The only thing that remains is getting the compile tests that you added working. Once that is done, we can merge!

rachitnigam:
This is ready to merge!

primitives/reshape.fil
```diff
@@ -67,21 +74,31 @@ comp Serialize[#W, #N, #B, ?#C=1, ?#H=1]<G: #C*(#N-1)+#H>(
 // In general, you probably don't want to use the #C and #H parameters at
 // all. They are most useful when the deserializer is consuming inputs from
 // a component that is not fully pipelined.
-comp Deserialize[#W, #N, #B, ?#C=1, ?#H=1]<G: #C*#N+#H-2>(
+comp Deserialize[#W, #N, #B, ?#C=1, ?#H=1]<G: #End-1>(
```
rachitnigam, line 77: The usage of `let` bindings made it obvious that we were holding onto the signals longer than we needed to!

src/ir/from_ast/astconv.rs
```diff
@@ -57,6 +56,18 @@ impl<'prog> BuildCtx<'prog> {
...
             )),
         };
+
+        // Extend the binding with the let-bound parameters in the signature
+        // Bindings can mention previous bindings so we add bindings as we
+        // go along.
+        comp.sig_binding.iter().for_each(
+            |ast::ParamBind { param, default }| {
+                let e = default.clone().unwrap().resolve(&binding);
+                binding.extend([(param.copy(), e)])
```
gabizon103, line 66: I will probably see as I continue reading, but where do we exclude the sig-bound parameters from the instance's parameter list? Because if I understand this correctly, by extending the binding that we use to make the `ir::Instance`, there will be parameters there that can't be passed in (causing the errors we saw before)?

src/ir/from_ast/build_ctx.rs
```diff
@@ -62,11 +62,15 @@ pub(super) struct BuildCtx<'prog> {
     sigs: &'prog SigMap,
 
     // Mapping from names to IR nodes.
-    event_map: ScopeMap<ir::Event>,
-    inst_map: ScopeMap<ir::Instance>,
-    inv_map: ScopeMap<ir::Invoke>,
-    port_map: ScopeMap<ir::Port, InvPort>,
-    param_map: ScopeMap<ir::Param>,
+    event_map: ScopeMap<ir::EventIdx>,
```
gabizon103, line 65: Question about data structures similar to ScopeMap (where we want to track a mapping to and/or from IR structures): should they take something like `Event` (and internally use `Idx<Event>`, which is what `Foreign` does) as a type or `EventIdx` (as done here)? I've seen it both ways throughout the code, and it might be good to make a consistent decision on this?

gabizon103 (approved):
I think this is good! Left small comments. I like making Filament look nicer!

src/expr_simplifier.rs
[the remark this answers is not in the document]
rachitnigam: Yup! I don't know why this file was committed in the repo anyways. I agree about the scope creep but this PR ended up solving a bunch of unrelated issues 😬

src/ir/from_ast/build_ctx.rs
rachitnigam: Yeah, this was how the interface originally worked by I ended up changing it to be generic scope tracking doesn't seem like it has anything to do with `Idx`. We should probably extract this into a utilities file somewhere instead of keeping it here since it can be generally useful.

src/ir/from_ast/astconv.rs
rachitnigam: So the `ir::Instance` object above takes the normal `binding` before we extend it to include the `sig_binding` stuff. This means that the instance gets parameter list we want while the bindings we store track the extra sig-bound parameters.

PR #302, opened 2023-09-02
Implement existentially quantified parameters

### LINKED ISSUE #121: Existentially quantified parameters
Parameters in Filament are universally quantified. For example, for the following module:
```rust
comp Foo[#W]() -> () where #W >= 32 { ... }
```
The parameter `#W` is universally quantified: forall values of `#W` that respect the constraints, we generate a component that is well-pipelined. Such universal quantification allows the calling context to decide how to use the module by providing a concrete instantiation for `#W`. In return, the calling context must prove that the binding for `#W` satisfies the constraint.

Existential types are the dual of universally quantified types. Instead of the calling context deciding how to use the module, the module tells the context how it's allowed to use it. For example:
```
comp ShiftSome[exists #L]<G: 1>(@[G, G+1] in: 32) -> (@[G+#L, G+#L+1] out: 32) 
```
This module delays the signal by some amount `#L` that is unknown to the calling context. Instead, when the calling context instantiates a shift circuit, it must use the specific instance of `#L` bound to that `ShiftSome` component:
```rust
comp main<G: 1>(... in: 32) -> () {
  S := new ShiftSome; // We don't provide a parameter here
  s0 := S<G>(in);
  mux := new Mux[32]<G+S.#L>(sel, s0.out, ...); // Scheduling must use S.#L to correctly use the signal s0.out
}
```

Note that when scheduling this program, we have to use `S.#L` to schedule things. `S.#L` is unknown when we write the program but will turned into a concrete value during compilation. While all invocations of `S` will get the same constant, different instantiations of `ShiftSome` might get different values of `#L`. This is important because it opens the door to something really cool–**per-instance retiming**.

Constraints on existential parameters must be satisfied by the body of the component defining them and can be assumed by the calling context.

## Compilation

The compilation strategy for existentials involves turning them into universally quantified types. This has the effect of turning the quantifiers "inside out". For each instance, we decide (somehow), the binding for each existential variable and replace all instantiations to use those constants. For example, in the above program, if we decide `#L=10` is a good binding for the variable, we change:
```
S := new ShiftSome;
```
Into:
```
S := new ShiftSome[10]
```
And make the signature of the component take to have `#L` be universally quantified parameter.

This is not fully developed yet but I think it's a good starting point.

rachitnigam:
Fixes #121.


rachitnigam:
Okay, basic parsing and astconv conversion is working. Next step is to separate the components into `assume` and `assert`. Previously, all component constraints were `assert`ed into the calling component by now some of them should be `assume`d. This is because constraints over existentially quantified parameters are supplied by instances instead of being proven by the caller.

After that, we need to figure out what to do with monomorphization. For now, maybe I'll require that all existential parameters have defined values in the program and eventually remove that constraint in a future PR when we want to be able to do retiming.

rachitnigam:
Okay, this PR now implements a basic AST to IR lowering for existential types. Monomorphization crashes when it sees existential parameters. A followup PR will build the flow that allows compilation of programs with existential types.

rachitnigam:
This should probably not be reviewed till #304 and #305 are merged. Currently, the diff on this is pretty big because it includes those changes.

rachitnigam:
Okay, this is ready to review now!

Issue #308, opened 2023-09-04
Functions on parameters using existential parameters 


UnsignedByte:
@rachitnigam and I discussed this in a meeting, but the addition of existential parameters by #121 enables creating components used only for operating on parameters.

For example, we can now do the following to implement `pow2`:
```
comp Pow2[N]<'G: 1>() -> () with {
    exists K;
} where K > 0, N >= 0 {
    if N == 0 {
        exists K = 1;
    } else {
        P := new Pow2[N-1]<'G>();
        exists K = P.K;
    }
}
```

This is a pretty cool observation as it allows us to do arbitrarily complex computation on parameters without making the compiler any more complicated, but still has some caveats:
* The assumptions necessary for type checking are currently not added using this, so such an implementation cannot be used somewhere that needs to be validated at that point in compilation (I.E. `ReduceAdd` or equivalent).
* These functions are components, so we cannot use them simply as functions on parameters, and can only define them in component bodies (though the existential parameters can likely be accessed outside of that scope, @rachitnigam correct me if I am wrong here).

This also allows us to use the constraints on `K` to generate assumptions in our parent component, which is also a nice observation (although `assume` already exists, so this may be less important). This is not really an issue, but just recording some of the possibilities with existential parameters, and maybe we could add some syntactic sugaring to this to make some cool features in the future.

rachitnigam:
The code in the example is not quite right. If we want to compute K = 2^N, then we have to do:
```
comp Pow2[N]<'G: 1>() -> () with {
    exists K where K > 0;
} N >= 0 {
    if N == 0 {
        exists K = 1;
    } else {
        P := new Pow2[N-1]<'G>();
        exists K = 2 * P.K;
    }
}
```

Apart from that, yeah, this is a cool capability. Essentially, forall parameters act as inputs to components and existentials are "outputs". The notable thing about this component is that it doesn't actually do anything at runtime; it's more like a "type-level" function whose only goal is to produce a new type based on a given type.

The reason this encoding is interesting (and tractable) is because while the computations on parameters can be arbitrary, the type system doesn't know about it. That is, even though, mathematically, we have the constraint that `K == 2^N`, the only fact exposed to the type system is that `K > 0` so users of this parameter can only rely on this knowledge.

rachitnigam:
[...]

PR #314, opened 2023-09-05
Monomorphize existential parameters

### LINKED ISSUE #310: Monomorphizing existentially quantified parameters
The newly added existentially quantified parameters cannot be compiled through to Calyx programs yet.
The problem is that monomorphization doesn't know how to deal with these parameters and I have realized that it's going to be a little trickier than I originally thought.

Consider the following program:
```
comp Baz[A]<'G: 1>() -> () with {
    exists E where E > A;
} {
    // This component constructs E using A
    exists E = A + 1;
}

comp Foo[A]<'G: 1>() -> () with {
    exists E where E > A;
} {
    BZ0 := new Baz[A];
    BZ1 := new Baz[A];
    // To return the value for this binding, we must monomorphize BZ0 and BZ1
    exists E = BZ0::E + BZ1::E;
}

comp main<'G: 1>() -> () {
    F0 := new Foo[10];
    // Monomorphizing this requires monomorphizing F0 first;
    F1 := new Foo[F0::E];
}
```

In this program, we have the following dependencies during monomorphization:
* `main` -> `F0`, `F1`
* `F1` -> `F0`
* `Foo` -> `BZ0`, `BZ1` ...

The second edge shows the source of the trickiness: monomorphization can depend on particular *instances* being handled instead of just components;
in order to know the value of `F0::E`, we need to monomorphize `F0` before `F1`.

Furthermore, we cannot just monomorphize the signature of `Foo` to get the value of `F0::E` since the parameter itself depends on `Baz`'s existentially quantified parameter, which requires us to evaluate the body of `Foo`.

One way to think about this trickiness is that normal parameters are "inputs" to component and existentially quantified parameters are "output" and components act as "functions" over these parameters.
In order to compute the right outputs, we need to completely evaluate all the component instantiations in a component.

rachitnigam:
Fixes #310.


rachitnigam:
Okay, looks like next step is to track `Foreign<ir::ParamIdx>` in the `ParamOwner` stuff so that we can correctly monomorphize things.

rachitnigam:
Okay, this works now!!

fil-ir/src/from_ast/sig_map.rs
```diff
@@ -1,49 +1,44 @@
...
 ///
 /// A signature defines the ports which are added to the component instantiating
 /// the signature.
 pub struct Sig {
     pub idx: ir::CompIdx,
-    pub events: Vec<ir::EventIdx>,
-    pub inputs: Vec<ir::PortIdx>,
-    pub outputs: Vec<ir::PortIdx>,
+    /// Mapping from source representation of input port to its IR index
+    pub inputs: Vec<(ast::Loc<ast::PortDef>, ir::PortIdx)>,
```
rachitnigam, line 16: The previous encoding relied on `inputs` containing the input ports in the same order as `raw_inputs`. This makes that dependency explicit.

src/ir_passes/mono/monosig.rs
```diff
@@ -222,25 +232,29 @@ impl MonoSig {
...
+                ir::ParamOwner::Instance { inst, .. } => format!(
                     "parameter defined by instance `{}'",
                     ul.display(inst.ul())
                 ),
                 ir::ParamOwner::Exists => {
-                    "existentially quantified parameter".to_string()
+                    unreachable!(
+                        "existential parameter `{}' occurred in a use location",
+                        p_rep
+                    )
```
UnsignedByte, line 251: Correct me if I'm wrong, but this implies that we can't use an existentially quantified parameters in any expressions?

UnsignedByte: Should we not just use `get_exist_val` like we do elsewhere?

rachitnigam: We add the existential parameter to the binding when monomorphize the corresponding instance:
```
A := new Foo;  // A::Out binding added here
B := new Bar[A::Out];
```

tests/compile/exist.fil
```diff
@@ -1,41 +1,49 @@
...
+    M1 := new Mul[32, 3];
     mul1 := M1<'G+M0::L>(mul0.out, in1);
 
+    // Next, test a reuse of the same instance.
+    mul2 := M1<'G+M0::L+M1::L>(mul1.out, in2);
+
     // This module defines the (hidden) relationship that the latency is the sum of the
     // the latencies of the two instances.
-    exists L = M0::L + M1::L;
+    exists L = M0::L + 2*M1::L;
```
rachitnigam, line 45: Oh awesome!

tests/check/exist-rec.fil
```diff
@@ -0,0 +1,17 @@
+import "primitives/core.fil";
+
+comp Pow[I]<'G:1>() -> () with {
```
rachitnigam, line 3: @UnsignedByte this example from #308 and it works!

fil-ir/src/validate.rs
gabizon103, line 1: General comment about the changes to `Validate`: what is the rationale for the things that were/weren't removed? I think the pass is in a weird spot where it isn't as robust as it could/should be, which is likely because it doesn't get used that much (especially since we now have the safer wrappers for mono). What do we want to do with it?

src/ir_passes/mono/monodeferred.rs
```diff
@@ -228,13 +269,16 @@ impl<'a, 'pass: 'a> MonoDeferred<'a, 'pass> {
...
             for cmd in body.iter() {
                 let cmd = self.command(cmd);
                 self.monosig.base.extend_cmds(cmd);
             }
-            // Pop all the let bindings
-            self.monosig.binding.pop(); // pop the index
+            // Remove all the bindings added in this scope including the index
+            self.monosig
+                .binding
+                .pop_n(self.monosig.binding.len() - orig_l);
```
gabizon103, line 281: Oh, didn't think of this! Cool!

src/ir_passes/mono/monomorphize.rs
```diff
@@ -40,14 +20,11 @@ pub struct Monomorphize<'a> {
...
     /// Instances that have already been processed. Tracks the name of the generated component
     pub processed: HashMap<CompKey, Base<ir::Component>>,
-    /// Instances that need to be generated
-    pub queue: LinkedHashMap<CompKey, (Base<ir::Component>, MonoSig)>,
 
     /// Mapping from old ports to new ports, for resolving Foreigns
-    pub port_map: HashMap<PortKey, Base<ir::Port>>,
-    /// Mapping from old events to new events, for resolving Foreigns
-    pub event_map: HashMap<EventKey, Base<ir::Event>>,
+    inst_info: HashMap<CompKey, InstanceInfo>,
```
gabizon103, line 25: comment?

src/ir_passes/mono/monomorphize.rs
```diff
@@ -88,19 +72,26 @@ impl<'ctx> Monomorphize<'ctx> {
 
         // If we've already processed this or queued this for processing, return the component
         if let Some(&name) = self.processed.get(&key) {
-            return (name, vec![]);
+            return name;
         }
 
-        if let Some((name, _)) = self.queue.get(&key) {
```
gabizon103, line 94: Can we get a comment somewhere here, or in the PR description maybe about how the structure of the pass has changed, for me to understand better and also so we have it for the future? I understand why it had to be changed from #310, but maybe documenting the change somewhere would be helpful

gabizon103 (approved):
Super cool! and looks good! This might be just a me thing, but how do we feel about the `exists <param> = <expr>` syntax? would a keyword like `bind` make more sense instead of `exists` here? Feel free to keep `exists`, but this was just a thought

fil-ir/src/from_ast/sig_map.rs
```diff
@@ -1,49 +1,44 @@
...
-use super::BuildRes;
-
 #[derive(Clone)]
 /// The signature of component.
 ///
 /// A signature defines the ports which are added to the component instantiating
 /// the signature.
 pub struct Sig {
     pub idx: ir::CompIdx,
-    pub events: Vec<ir::EventIdx>,
```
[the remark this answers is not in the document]
rachitnigam: It seems we never used the events? Do you remember what this was supposed to be used for?

rachitnigam:
I must confess that I’m not a huge fan of the exists binding either and am up for other suggestions for it.

src/ir_passes/mono/monomorphize.rs
rachitnigam: Good point! I can add something to the description of the monomorphize pass but I also think we should write a doc page on filamenthdl.com/docs that describes more internal details of the compiler so that it's easier to remember in the future.

src/ir_passes/mono/monomorphize.rs
rachitnigam: There is a comment about `inst_info` already. Do you mean something else?

src/ir_passes/mono/monomorphize.rs
rachitnigam: Done!

fil-ir/src/validate.rs
rachitnigam: Yeah, the problem with `Validate` is that it had a lot of infinite loops of the form where `expr` would call `param` which would call `expr`. The basic idea behind the changes is that we don't check any `Interned` value because we assume them to be valid and they cannot be removed because `Interned` doesn't implement `MutCtx`. Values in `IndexStore` need to be checked that they are valid.

We still don't have a pass that actually deletes things from `IndexStore` so we should implement something like that to exercise that capability.

PR #317, opened 2023-09-09
Expands the primitives library a little

### LINKED ISSUE #316: Primitive `Slice` used ambiguously
Slice is implemented as follows:
```v
module Slice #(
  parameter IN_WIDTH = 32,
  parameter MSB = 31,
  parameter LSB = 0,
  parameter OUT_WIDTH = 32
) (
  input wire logic [IN_WIDTH-1:0] in,
  output wire logic [OUT_WIDTH-1:0] out
);
  assign out = in[MSB:LSB];
endmodule
```
which makes `MSB` inclusive, and thus we theoretically want `OUT_WIDTH = MSB - LSB + 1` as a constraint. However, within `primitives/math.fil` component `Id_Init` declares a 
```rust
q_slice := new Slice[Ext, W, 0, W]<'G>(c_ext.out);
```
which would violate this constraint (I'm assuming `W` is meant to be non-inclusive here).
Which should we choose to use? It might be worth doing non-inclusive slicing as it would match with bundle indexing/range generation, etc. I plan to change the last parameter `OUT_WIDTH` to be optional with a default, and another small question would be whether to swap the order of the `MSB` and `LSB` parameters maybe.

UnsignedByte:
Adds `ConcatBundle` and `SplitWire`, combinational components that combine bundles into a single wire and to split a single wire into a bundle.

Also adds `FNeg` and `FSub`, components to negate and subtract floating point numbers.

Will fix #316, which should remove some ambiguity in the way we use `Slice` at the moment.


primitives/comb.fil
```diff
@@ -177,4 +178,42 @@ extern "comb.sv" {
    ) -> (
       out: ['G, 'L] OUT_WIDTH
    ) where 'L > 'G, IN_WIDTH > 0, OUT_WIDTH > 0;
+}
+
+// Reduces a bundle of N wires of size W to a single wire of size W*N
+// {010, 101, 111} -> 010101111
+comp ConcatBundle[W, N]<'G: 1>(
```
rachitnigam, line 185: Let's put these in a different file. In general, I want primitives files that expose externs to only contain the externs defined in the Verilog file.

primitives/comb.fil
```diff
@@ -177,4 +178,42 @@ extern "comb.sv" {
...
+   in[N]: ['G, 'G + 1] W
+) -> (
+   out: ['G, 'G + 1] W*N
+) where N > 0, W > 0 {
+   if N == 1 {
+      out = in{0};
+   } else {
+      // recursively reduce
+      red := new ConcatBundle[W, N-1]<'G>(in{1..N});
+      c := new Concat[W, W*(N-1)]<'G>(in{0}, red.out);
```
rachitnigam, line 195: Awesome!

primitives/comb.fil
```diff
@@ -177,4 +178,42 @@ extern "comb.sv" {
...
+      out = c.out;
+   }
+}
+
+// Splits a wire of size W*N into a bundle of N wires of size W
+// 010101111 -> {010, 101, 111}
+comp SplitWire[W, N]<'G: 1>(
```
rachitnigam, line 202: Heck yes! This was exactly what I was thinking about during our conversation last week.

UnsignedByte:
@rachitnigam I moved the components to `reshape` if that's alright? If not, I can move them to their own file altogether.

rachitnigam:
Sounds good!

PR #325, opened 2023-09-14
Expand built-in functions

UnsignedByte:
Adds a bunch of math functions (`sin`, `cos`, etc) to the built-in function set (which currently only has `log2` and `pow2`.

This PR now adds a small type system within expressions: instead of Concrete exprs working on `u64` values only, instead they now store an `ast::Concrete` that is either a `Concrete::Uint(u64)` or `Concrete::Float(f64)`. Currently, during evaluation:
* `u64` automatically casts to `f64` (@rachitnigam we might want to discuss this) to prevent having separate implementations for the same functions
* `f64` needs to be cast to `u64` using functions like `floor`, `ceil` (so that it is explicit how they are converted).

~~Trigonometric functions will likely work as follows (@rachitnigam open to discussion on this, but I can't really think of a better solution atm):~~
* ~~Each function will take two arguments, `num` and `den`, the numerator and denominator of a fraction, and performs the operation `sin(num/den * 2pi)` (example for `sin` only).~~
* ~~Because functions can only operate on parameters (internally `u64`), we will have variations of each function for `f32` and `f64`, `sin32` and `sin64` that return the bits of the `f32` and `f64` functions interpreted as an integer. This way, filament components such as `Fadd`, etc can operate on these functions.~~


fil-ast/src/expr.rs
```diff
@@ -1,5 +1,170 @@
...
+        }
+    }
+}
+
+impl TryFrom<Concrete> for f64 {
+    type Error = Error;
+
+    fn try_from(value: Concrete) -> Result<Self, Self::Error> {
+        match value {
+            Concrete::UInt(n) => Ok(n as f64),
```
UnsignedByte, line 67: @rachitnigam We should discuss whether uints auto-casting to floats should be allowed. This simplifies implementation pretty significantly because functions can take floats and they'll automatically evaluate for integers as well, but it would be (insignificantly) lossy for very large `u64`s

src/ir_passes/discharge.rs
```diff
@@ -170,18 +170,21 @@ impl Discharge {
...
+                    .sol
+                    .declare_fun(stringify!($name).to_lowercase(), vec![$($args),*], $out)
+                    .unwrap());
+            }
+        );
+
+        sol_fn!(Pow2(is) -> is);
+        sol_fn!(CLog2(is) -> is);
+        sol_fn!(Pow(is, is) -> is);
+        sol_fn!(CLog(is, is) -> is);
```
UnsignedByte, line 187: Another issue here: Because the solver can only work on integer values, if we used something like `ceil` in any parameter expression that eventually is used in typechecking, it can lead to broken solver behavior (I think). @rachitnigam is there a way to get around this, or do we need to instead add some checks in the `Validate` pass on whether float-based computations are used anywhere in expressions used by the typechecker?

fil-ast/src/expr.rs
UnsignedByte: FWIW, this shouldn't change behavior at all currently, as internally `u64` concrete values were already cast to `f64` to perform `log` (and that is legitimately probably the best way to do it).

fil-ast/src/expr.rs
```diff
@@ -25,50 +190,55 @@ impl std::fmt::Display for Op {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd)]
 /// A unary uninterpreted function over integers.
-pub enum UnFn {
-    /// The `pow2` function
+pub enum Fn {
```
UnsignedByte, line 193: `ast::Fn` is currently used in a lot of `ir` locations, but we might want to create an `ir::Fn` version that simplifies a lot of the functions away. For example, `Log2` and `Pow2` can be replaced by `log(2, ...)` and `pow(2, ...)`, and similarly `CLog` can be replaced with `ceil(log(...))` which could be done in `astconv`.

UnsignedByte:
Will revert this branch to [62451fd](https://github.com/cucapra/filament/pull/325/commits/62451fd56eec207d19d867958aeee15d1da4a66b) and merge that without the new type system instead, because of the issues mentioned in #327.

UnsignedByte:
e

PR #326, opened 2023-09-15
Rewrite Aetherling evaluation using existential types

rachitnigam:
This defines an interface for all the `Conv2d` modules generated by the Aetherling compiler using existential types. The upshot is that we can abstract the entire design space of the Aetherling modules by simply defining existential types and use that as a stable interface for the host code.

This capability is going to be key in arguing that parameteric Filament is a powerful tool for integrating code generated from various ADLs and RTL in a safe manner (in addition to enabling safe design).


rachitnigam:
Okay, this is ready to be merged. Instead of one interface, I ended up with two. The "fast" interface can wrap all the pipelined modules while the "slow" interface wraps the slow modules. Generalizing the two seems to involve a lot more tension and feels like unnecessary parameterization. I've run into a bunch of confusing error when building this which is definitely good because it'll help us refine the experience with existential parameters.

PR #328, opened 2023-09-19
Fixes bugs in FP-Add and Iter-Div and add Edge Test cases

### LINKED ISSUE #329: Inconsistent behavior with Iterative Divider
The `verilog` implementation of the iterative divider returns `0` upon the division `0/0`, while all the other implementations return `255` (max value).

Placing the following test in `edge.json` (provided #328 is merged) will show off this behavior:
```json
{
    "left": [0],
    "right": [0]
}
```

UnsignedByte:
Fixes a bug in renormalization when all mantissa bits are `0`.

Also fixes inconsistent behavior between verilog and filament implementations with the iterative divider when performing `0/0`, and a bug that made the verilog implementation non-pipelineable. (Closes #329)

Also adds testing capability for edge cases with FP library and iterative divider.


tests/iterative-divider/iter-div.sv
```diff
@@ -19,7 +19,11 @@ module IterDiv #(
     logic start, running, finished, dividend_is_zero;
 
     assign start = go && !running;
-    assign dividend_is_zero = start && left == 0;
+    // NOTE: 0 / 0 returns 0 due to this optimization.
+    // This also means that the delay of the component is non-constant.
+    // Otherwise, would return `(1 << W) - 1` in `W` cycles
+    // assign dividend_is_zero = start && left == 0;
+    assign dividend_is_zero = 0;
```
UnsignedByte, line 26: #329 Expands on why this change was made.

tests/iterative-divider/iter-div.sv
```diff
@@ -19,7 +19,12 @@ module IterDiv #(
     logic start, running, finished, dividend_is_zero;
 
     assign start = go && !running;
-    assign dividend_is_zero = start && left == 0;
+    // BUGFIX: 0 / 0 returns 0 due to this optimization.
+    // Also, as this causes values to be returned immediately,
+    // It makes the implementation non-representable by filament.
+    // Otherwise, would return `(1 << W) - 1` in `W` cycles
+    // assign dividend_is_zero = start && left == 0;
+    assign dividend_is_zero = 0;
```
UnsignedByte, line 27: #329 Expands on why this change was made.

rachitnigam: This makes a lot of sense to me!

rachitnigam:
Changes seem fine but another reasonable solution is not testing division with `0`. Since we've already put in the effort here, might as well commit it.

UnsignedByte:
This might have ended up being more work, because with a `1/255^2` chance, the random tester could have generated two zeros, and we'd have to have some code to allow for excluding certain pairs when randomized testing.

PR #335, opened 2023-09-22
Fix more floating point bugs.

### LINKED ISSUE #334: Floating point multiplier bug in all implementations.
Pretty poetic that #331 was just mentioned, but there's a bug in floating point multiplication when all mantissa bits are `0`. This is because the multiplication algorithm forgets to extend the mantissa to include the always-`1` extra bit, so loses that information. This is a bug in every single implementation, and thus was never caught.

UnsignedByte:
Fixes #334.

Removes the `zero` check for whether the mantissa equals zero.

Further changes renormalization for fp-add.


tests/floating-point/mul/fp-mult-nopipe.sv
```diff
@@ -28,15 +28,12 @@ module FP_Mult_NoPipe (
   assign round = |product_normalised[22:0];  // Last 22 bits are ORed for rounding off purpose
   assign product_mantissa = product_normalised[46:24] + (product_normalised[23] & round); 					// Mantissa
 
-  assign zero = exception | (product_mantissa == 23'd0) ;
-
```
UnsignedByte, line 32: @rachitnigam does this look right? What was the zero check initially for? Not sure whether I might be causing another bug by fixing this one.

rachitnigam: Sorry, I didn't write this code so I can't answer here. The Berkeley HardFloat implementation might provide a better guideline on things like this.

UnsignedByte:
Closing this as it is horribly behind.

PR #340, opened 2023-09-29
Disallow use of existential parameters in the body

### LINKED ISSUE #336: Confusing error message when existential parameter used in the body
When trying to compile the following program:
```
comp Foo<'G:1>() -> () with {
    exists N;
} {
    // Bundle uses N which will not be bound in time during monomorphization.
    bundle a[N]: ['G, 'G+1] 32;
    exists N = 10;
}

comp main<'G:1>() -> () {
    F := new Foo;
}
```

The compiler says:
```
rachitnigam ~/git/filament (aetherling-existential !?$)
% cargo run -q -- exist-bad-err.fil
thread 'main' panicked at 'internal error: entered unreachable code: existential parameter `N' occurred in a use location', src/ir_passes/mono/monosig.rs:235:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The actual problem in the program is use of `N` in `bundle a[N]`. The problem is that when running the program, `N` is not yet bound to a value because we have not evaluated the `exists` binding. In general, existential parameters cannot be used in the body because of this problem. Instead, the correct program should just say `bundle a[N]`.

Our well-formedness pass should emit an error to this effect.

rachitnigam:
Fixes #336.

### DIFF
--- a/fil-ir/src/from_ast/build_ctx.rs
+++ b/fil-ir/src/from_ast/build_ctx.rs
@@ -1,6 +1,6 @@
 use super::sig_map::Sig;
 use super::{BuildRes, ScopeMap, SigMap};
-use crate::{self as ir, Ctx, DenseIndexInfo, PortIdx};
+use crate::{self as ir, Ctx, DenseIndexInfo, ParamOwner, PortIdx};
 use fil_ast as ast;
 use fil_utils::{self as utils, Error, Id};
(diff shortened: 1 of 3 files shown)


rachitnigam:
This didn't work out. We'll try again next time. C'est la vie.

PR #344, opened 2023-10-06
Combinational Pease FFT

UnsignedByte:
Builds an FFT using the pease dataflow. This will be super useful for pipelining, as every stage performs exactly the same computation, making parallelism much easier.

Also restructures the FFTs to prepare for the pipelined version.


rachitnigam:
Any references for "Pease dataflow"? I must confess I've not come across this before

UnsignedByte:
Page 6 of spiral's paper describes it [here](https://www.spiral.net/doc/papers/fft-enc11.pdf); It's pretty much an FFT where the crossbar (reordering of inputs) is the same every cycle. This makes implementation a lot easier as every stage is identical, so we just have to repeat the same computation `NStages` times to complete the FFT.

UnsignedByte:
Surprising how easy this was to implement, looks like the testing is going well! I do think my logic for `pease` might be a little off so I might increase the testing size from 8 to 16 points.

UnsignedByte:
Also a surprisingly easy change, and the FFT seems to be working as-is!

rachitnigam:
Woohoo! I'm going to credit your SWE skillz and Filament for the ease of extension!! One thing I'd be curious about at some point is your reflections from implementing the whole FFT project and which parts felt easy. It would be particularly interesting because you've been building it in Verilog as well (I assume for C2S2).

tests/floating-point/fft/fft-comb-pease.fil
```diff
@@ -0,0 +1,124 @@
...
+
+    let data_offset = NPoints * stage * 2;
+
+    // temporary data storage before stride permutation
+    bundle tmp_data[NPoints*2]: ['G, 'G+1] W;
+
+    // perform the butterflies
+    for butterfly_i in 0..pow2(NStages-1) {
+      // needed to speed up the solver
+      assume butterfly_i < NPoints/2;
```
rachitnigam, line 51: Out of curiosity: how did you figure out that this was the thing that was needed?

tests/floating-point/fft/fft-comb-pease.fil
```diff
@@ -0,0 +1,124 @@
...
+    data{data_offset+NPoints*2..data_offset+NPoints*4} = stride_perm.out{0..NPoints*2};
+  }
+
+  // sp := new StridePermutation[NStages, W]<'G>(in{0..NPoints*2});
+  // out{0..NPoints*2} = sp.out{0..NPoints*2};
+  // twid := new TwiddleStage[NStages, 0, W]<'G>(twiddle{0..NPoints});
+  // out{0..NPoints} = twid.out{0..NPoints};
+  // out{NPoints..NPoints*2} = twiddle{0..NPoints};
+
+  // out{0..NPoints*2} = data{NPoints*1*2..NPoints*1*2+NPoints*2};
```
rachitnigam, line 75: Any reason to keep this commented out lines?

tests/floating-point/fft/fft-comb-pease.fil
```diff
@@ -0,0 +1,124 @@
...
+  out{0..NPoints*2} = data{NPoints*NStages*2..NPoints*NStages*2+NPoints*2};
```
rachitnigam, line 76: Would be worth naming these complex expressions involving parameters: `NPoints*2` and `NPoints*NStages`

tests/floating-point/fft/harness.fil
```diff
@@ -23,49 +23,81 @@ extern "spiral-fft.v" {
...
+    X22: ['G, 'G+1] 32,
+    X23: ['G, 'G+1] 32,
+    X24: ['G, 'G+1] 32,
+    X25: ['G, 'G+1] 32,
+    X26: ['G, 'G+1] 32,
+    X27: ['G, 'G+1] 32,
+    X28: ['G, 'G+1] 32,
+    X29: ['G, 'G+1] 32,
+    X30: ['G, 'G+1] 32,
+    X31: ['G, 'G+1] 32
```
rachitnigam, line 41: These kinds of interfaces show up in Aetherling too. I would be interesting to come up with some predictable lowering scheme where you can write `X[32]` (bundle syntax) for external components or `main` and have the guarantee that it will generate an interface with names `X0`, `X1`, ...

tests/floating-point/fft/harness.fil
```diff
@@ -103,7 +151,23 @@ comp main<'G: 1>(
...
+  spiral_bundle{22} = spdft.Y22;
+  spiral_bundle{23} = spdft.Y23;
+  spiral_bundle{24} = spdft.Y24;
+  spiral_bundle{25} = spdft.Y25;
+  spiral_bundle{26} = spdft.Y26;
+  spiral_bundle{27} = spdft.Y27;
+  spiral_bundle{28} = spdft.Y28;
+  spiral_bundle{29} = spdft.Y29;
+  spiral_bundle{30} = spdft.Y30;
+  spiral_bundle{31} = spdft.Y31;
```
rachitnigam, line 169: Another reason to support the whole "predictable bundle lowering" for external interfaces is so that we don't have to write these unrolled assignments by hand (I do this for Aetherling too)

rachitnigam (approved):
Minor comments; resolve and merge at your discretion! Awesome work on this!! I'm super excited to see that speed at which you've been able to make these changes and excited to see how we'll compare to Spiral generated designs.

tests/floating-point/fft/fft-comb-pease.fil
UnsignedByte: I realized after commenting out other sections of the code that this was nearly freezing `cvc5`, and I added this to make it work. I think it's actually not necessary now because we have that `pow2(NStages-1) == NPoints/2` further above.

tests/floating-point/fft/fft-comb-pease.fil
UnsignedByte: Nope removed them!

tests/floating-point/fft/harness.fil
UnsignedByte: Yeah, that would be super nice! Maybe something to implement as a nice feature for externs.

tests/floating-point/fft/fft-comb-pease.fil
rachitnigam: Cvc5 has been consistently slower so we should replace it with z3 as the default solver

PR #345, opened 2023-10-07
move assertion generation out of astconv

rachitnigam:


rachitnigam:
This will require a bunch of work to merge unfortunately. The basic problem is that we need to have a nice way to rewrite the assertions from the signature of a component into a `PropIdx` of the current component. The big problem is that there is no good way to call `ir::Subst` on `Idx` defined on a different component.

There is a more general question about how to support inlining of `ExprIdx` and `PropIdx` like things from a different component into the current component. In the future this would allow us to literally implement inlining of instances into a particular context.

rachitnigam:
This can't work till we have #371

PR #348, opened 2023-10-12
Created Pipelined Pease FFT

UnsignedByte:
Builds a pipelined PEASE FFT.


tests/floating-point/fft/implementations/fft-pipelined.fil
```diff
@@ -0,0 +1,149 @@
...
+      // concat into one wire for the register
+      cb := new ConcatBundle[W, 4*N]<'G + stage_time + (i + 1) * B::L>(b.out{0..4*N});
+      // register to hold values till output time
+      // TODO (edmund): This register can be reused per stage
+      reg := new Register[W*4*N]<'G + stage_time + (i + 1) * B::L, 'G + stage_time + stage_latency + 1>(cb.out);
+      // split back into bundle
+      sp := new SplitWire[W, 4*N]<'G + stage_time + stage_latency>(reg.out);
+
+      // Write to the next stage
+      data{start + NPoints*2..end + NPoints*2} = sp.out{0..4*N};
```
rachitnigam, line 66: Very surprisingly, commenting this line out makes the SMT solver terminate quickly. I've tried adding more assumptions like `end - start == 4*N` but that doesn't seem to help.

rachitnigam: Actually, maybe this is not that surprising give that `data` has a really complex liveness type and `sp` probably gets a really complicated one when `stage_time + stage_latency` is expanded.

rachitnigam: Okay, @UnsignedByte I've added a new assumption that seems to help. Note that only `z3` seems to terminate with these changes so cvc5 still won't work. The problem still seems to arise from the use of `/` which is maybe unsurprising. This makes a pretty strong case for #351.

UnsignedByte: Thanks! Still not passing on `z3` for me on the ssh machine, which is weird. I'll take another look.

tests/floating-point/fft/implementations/fft-pipelined.fil
```diff
@@ -43,20 +44,18 @@ comp PeaseFFT[NStages, N]<'G: L>(
...
+    twiddle_stage := new TwiddleStage[NStages, twiddle_i, W, latency]<'G>(twiddle.out{0..NPoints/2}{0..2});
+    twiddle_in{twiddle_i}{0..NPoints/2}{0..2} = twiddle_stage.out{0..NPoints/2}{0..2};
   }
 
   // Synchronized data (happens between each stage, inputs and outputs to stride permutation)
   // Separated by the input and output to and from the registers.
   // There are Nstages of these (one for input and one for output for each stage)
-  bundle sync_data_in[(NStages + 1) * NPoints * 2]: for<sdii> ['G + sdii/(NPoints*2) * stage_latency, 'G + sdii/(NPoints*2) * stage_latency + 1] W;
-  bundle sync_data_out[NStages * NPoints * 2]:  for<sdoi> ['G + (sdoi/(NPoints*2)+1) * stage_latency, 'G + (sdoi/(NPoints*2)+1) * stage_latency + 1] W;
+  bundle sync_data_in[NStages+1][NPoints][2]: for<sdii, _sdi, __sdi> ['G + sdii*stage_latency, 'G + sdii*stage_latency + 1] W;
```
rachitnigam, line 57: You don't need to bind unused indices. You can just do `for<sdii>` and ignore the other two.

UnsignedByte:
I think we have to swap to `z3` to make this pass - currently timing out due to discharge because I added more to the test harness.

UnsignedByte:
Tests pass now, LMK whether the stuff in the apps directory is organized well, and then we should be able to merge this.

apps/fft/src/fft.fil
```diff
@@ -0,0 +1,194 @@
...
+  // cvc5 stalls trying to prove this...
+  assume NPoints % 2 == 0;
+
+  assume pow2(NStages-1) == NPoints/2;
+  assume pow2(NStages+1) == NPoints*2;
+  assume pow2(NStages+2) == NPoints*4;
+
+  bitrev := new BitRev[NStages, W]<'G>(in{0..NPoints}{0..2});
+  // butterfly components
+  B := new Butterflies[N, W, E, M];
```
rachitnigam, line 37: Let's add borrow annotations to these so that we know that we have the right delays

apps/fft/src/butterfly.fil
```diff
@@ -0,0 +1,88 @@
...
+) with {
+  some L where L > 0;
+} where
+    W == E + M + 1,
+    E > 1,
+    M > 1,
+    L > 0
+{
+  // switches between the type of butterfly unit
+  let T = 1;
```
rachitnigam, line 19: We should eventually expose some way for Filament designs to access a global `config` object to do this kind of switching without needing to change the source code. Nothing actionable; just a passing thought.

apps/fft/src/butterfly.fil
```diff
@@ -0,0 +1,88 @@
...
+      FA := new FAdd[W, E, M];
+      FS := new FSub[W, E, M];
+      
+      // Perform w * in1
+      ac := FM1<'G>(in1{0}, twiddle{0}); // ac
+      bd := FM2<'G>(in1{1}, twiddle{1}); // bd
+      re := FS<'G>(ac.out, bd.out); // ac - bd
+      re_d1 := new Delay[W]<'G>(re.out); // delay by one cycle
+      re_d2 := new Delay[W]<'G+1>(re_d1.out); // delay to second cycle
+      re_d3 := new Delay[W]<'G+2>(re_d2.out); // delay to third cycle
```
rachitnigam, line 57: Replace with a `Shift[3]`?

apps/fft/src/butterfly.fil
```diff
@@ -0,0 +1,88 @@
...
+
+      ad := FM1<'G+1>(in1{0}, twiddle{1}); // ad
+      bc := FM2<'G+1>(in1{1}, twiddle{0}); // bc
+      im := FA<'G+1>(ad.out, bc.out); // ad + bc
+      im_d2 := new Delay[W]<'G+1>(im.out); // delay to second cycle
+      im_d3 := new Delay[W]<'G+2>(im_d2.out); // delay to third cycle
```
rachitnigam, line 63: Replace with a `Shift[2]`

apps/fft/src/butterfly.fil
```diff
@@ -0,0 +1,88 @@
...
+      // in0 - (w * in1)
+      out10 := FS<'G+2>(in0{0}, re_d2.out); // re
+      out11 := FS<'G+3>(in0{1}, im_d3.out); // im
+
+      out00_d3 := new Delay[W]<'G+2>(out00.out); // delay to third cycle
+      out00_d4 := new Delay[W]<'G+3>(out00_d3.out); // delay to fourth cycle
+      out01_d4 := new Delay[W]<'G+3>(out01.out); // delay to fourth cycle
+      out10_d3 := new Delay[W]<'G+2>(out10.out); // delay to third cycle
+      out10_d4 := new Delay[W]<'G+3>(out10_d3.out); // delay to fourth cycle
+      out11_d4 := new Delay[W]<'G+3>(out11.out); // delay to fourth cycle
```
rachitnigam, line 78: `Shift[5]`! Abstractions! Woo!

apps/fft/src/fft.fil
```diff
@@ -0,0 +1,187 @@
...
+  out[N*2][2]: ['G+L, 'G+L+1] W
+) with {
+  some L where L > 0;
+} where
+    N > 0,
+    W == E + M + 1,
+    E > 1,
+    M > 1
+{
+  Bx := new Butterfly[W, E, M];
```
rachitnigam, line 141: Let's document why we need to create this at all

apps/fft/src/fft.fil
```diff
@@ -0,0 +1,187 @@
...
+) where
+    N > 0,
+    W > 0,
+    L > 0
+{
+  bundle outb[N*4]: ['G, 'G+L] W;
+  for i in 0..N*2 {
+    outb{i*2} = in{i}{0};
+    outb{i*2+1} = in{i}{1};
+  }
```
rachitnigam, line 165: This looks like a common pattern that might be useful to put into reshape sometime: the ability to interleave two bundles into one. Unfortunately, this is not easily done because we want to abstract over the timing of the inputs and outputs and say something like:
```
comp Interleave[N, M]<'G:1, 'L:1>(
  in0[N]: ['G, 'L], in1[M]: ['G, 'L]
) -> (out[N+M]: ['G, 'L]) where 'L > 'G {...}
```
This won't work because we can't put ordering constraints on the events. It occurs to me that there are some components that have these constraints but no interface ports. These should be easier to support because we know that during compilation, we don't have to generate multiple FSMs.

apps/fft/src/fft.fil
```diff
@@ -0,0 +1,187 @@
...
+  
+  cb := new ConcatBundle[W, N*4, L]<'G>(outb{0..N*4});
+  
+  out = cb.out;
+}
+
+/// Splits output of the Butterflies component into a bundle
+comp SplitButterflyBundle[N, W, ?L=1]<'G: L>(
```
rachitnigam, line 173: This is the opposite operation: Partitioning. Also worth being in reshape library one day!

apps/fft/test/harness.fil
```diff
@@ -76,28 +74,92 @@ extern "spiral-fft.v" {
...
+  NStages > 0
 {
-  // split input into bundle
-  split := new SplitWire[32, 32]<'G>(in);
+  bundle input_bundle[NPoints][2]: ['G, 'G+1] 32;
+
+  for i in 0..NPoints {
+    input_bundle{i}{0} = in{2*i};
+    input_bundle{i}{1} = in{2*i+1};
+  }
```
rachitnigam, line 97: Another partition use here.

apps/fft/test/harness.fil
```diff
@@ -76,28 +74,92 @@ extern "spiral-fft.v" {
...
+  // hook up pipelined fft
+  FP := new PeaseFFT[NStages, N];
+  fp := FP<'G>(input_bundle{0..NPoints}{0..2});
+
+  bundle fp_output_bundle[NPoints*2]: ['G+FP::L, 'G+FP::L+1] 32;
+
+  for i in 0..NPoints {
+    fp_output_bundle{2*i} = fp.out{i}{0};
+    fp_output_bundle{2*i+1} = fp.out{i}{1};
+  }
```
rachitnigam, line 108: Another Interleave use here.

apps/fft/test/harness.fil
```diff
@@ -76,28 +74,92 @@ extern "spiral-fft.v" {
...
+  fil_pipelined_4: ['G+FP_4L, 'G+FP_4L+1] NPoints*2*32, // output from pipelined pease fft with 4 butterflies
+  fil_pipelined_8: ['G+FP_8L, 'G+FP_8L+1] NPoints*2*32, // output from pipelined pease fft with 8 butterflies
+) with {
+  let NStages = 4;
+  let NPoints = pow2(NStages);
+
+  some FP_8L where FP_8L > 0;
+  some FP_4L where FP_4L > 0;
+  some FP_2L where FP_2L > 0;
+  some FP_1L where FP_1L > 0;
```
rachitnigam, line 132: Love it!

primitives/signed.fil
```diff
@@ -8,17 +8,18 @@ extern "signed.sv" {
    ) where 'L > 'G, WIDTH > 0;
 }
 
-comp SignedConst[WIDTH, VALUE]<'G: 1>() -> (
-   out: ['G, 'G + 1] WIDTH,
+comp SignedConst[WIDTH, VALUE, ?D=1]<'G: D>() -> (
+   out: ['G, 'G+D] WIDTH,
```
rachitnigam, line 12: This component is another example of something that can have multiple events but doesn't need any compilation support for them.

rachitnigam (approved):
LGTM! Some refactoring suggestions and some flyby thoughts for future work. Awesome work on this! Next up, we should measure the resource usage of these designs and compare to Spiral!

PR #350, opened 2023-10-12
Add discharge pass flag

### LINKED ISSUE #349: Discharge flag to assert statements one by one
During the discharge pass, because assertions are sent to the solver all at once, if something is very slow it makes it difficult to figure out what is causing the solver to slow. The discharge pass already contains the capability to send statements one by one instead (as this is what allows us to generate diagnostics) so we should have a flag the lets us use this behavior by default for debugging purposes.

UnsignedByte:
Closes #349

### DIFF
--- a/src/ir_passes/discharge.rs
+++ b/src/ir_passes/discharge.rs
@@ -244,7 +244,7 @@ impl Discharge {
     }
 
     /// Check whether the proposition is valid.
-    /// Returns a set of assignments if the proposition is not valid.
+    /// Panics with a set of assignments if the proposition is not valid.
     fn check_valid(&mut self, fact: ir::Fact, ctx: &ir::Component) {
         let prop = fact.prop;
         #[allow(clippy::map_entry)]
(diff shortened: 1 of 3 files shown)


UnsignedByte:
Pretty simple to implement

src/ir_passes/discharge.rs
```diff
@@ -244,7 +244,7 @@ impl Discharge {
     }
 
     /// Check whether the proposition is valid.
-    /// Returns a set of assignments if the proposition is not valid.
+    /// Panics with a set of assignments if the proposition is not valid.
```
rachitnigam, line 247: Sorry, why will this panic?

rachitnigam (approved):
One question about the comment update but otherwise LGTM. Merge at discretion if there is no functional change in the implementation

src/ir_passes/discharge.rs
rachitnigam: I've fixed the comment and merging!

PR #352, opened 2023-10-13
Set up `.devcontainer`

UnsignedByte:
Sets up docker with vscode for filament.


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ghcr.io/cucapra/calyx:0.4.0
+
+WORKDIR /home
+ADD . filament
+# Build the compiler
+WORKDIR /home/filament
+RUN CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse cargo build --all
```
rachitnigam, line 7: Might be worth installing `z3` and `cvc5` and setting up the `fud` stage like we do in the github actions: https://github.com/cucapra/filament/blob/main/.github/workflows/test.yml#L11

rachitnigam (approved):
Generally LGTM!

Dockerfile
UnsignedByte, line 7: This uses the [sparse registry](https://blog.rust-lang.org/inside-rust/2023/01/30/cargo-sparse-protocol.html) protocol, mainly because on my machine downloading all of `crates.io` took something like 20 minutes which was not the greatest. Not sure whether we should keep this though, as (I think) the next time you run cargo build it'll try to download everything again.

UnsignedByte: Removed this for now.

PR #363, opened 2023-10-15
Opaque and transparent existential parameters

### LINKED ISSUE #354: Pure function interpretation of `comp` definitions over parameters
Consider the following program:
```
comp Foo[I](...) with {
  exists O;
} { ... }
comp Bar(...) {
  F0 = Foo[10];
  F1 = Foo[10];
  F2 = Foo[20];

  assert F0::O == F1::O;
  assert F0::O == F2::O;
}
```

The question which `assert` statements, if any, should be allowed to pass. Here are the possibilities:
1. **No `assert` passes**: This means that *all instances*, regardless of input parameters, can produce different values for existential parameters.
2. **The first `assert` passes**: This means that components are *pure functions* over parameters; the only thing that can influence the value of a parameter is input parameters. The body can of course use existential parameters defined by other subcomponents but those themselves can only be influenced by input parameters.
3. **Both `assert`s pass**: This means that there must have been some language-level mechanism to declare that the existential parameter `O` has the same value for all instances of the module. In this case, the parameter *must be a constant* value and a more appropriate thing to say would be that existential parameter is defined over the component definition instead of an instance (and use the syntax `Foo::O` instead of `F0::O`).

I think the current status, which is (1), is not tenable because it doesn't allow us to build module hierarchies where we instantiate `N` copies of the same implementation, chain them together, and prove that the latency is `N*L`. We either need (2) or (3). 

The nice thing about (2) is that we can encode it really efficiently in SMT by saying that each component is a function over parameters:
```
(define-fun (foo Int) Int)
(assert (= F0::L  (foo 10))
(assert (= F1::L  (foo 10))
(assert (= F2::L  (foo 10))
```

This would be sufficient to make the first assertion pass. The downside is that we lose the ability to do instance-specific parameterization but maybe that should be an opt-in feature.

CC @UnsignedByte @gabizon103 @sampsyo

rachitnigam:
Fixes #354. Implements the proposed syntax changes from the issue as well.

### DIFF
--- a/apps/aetherling/common/attempt.fil
+++ b/apps/aetherling/common/attempt.fil
@@ -24,10 +24,10 @@ comp Conv2dAbstract<'G: II>(
     out{0..C::N} = c.O{0..C::N};
 
     // Parameters are the same as the wrapped module
-    exists L = C::L;
-    exists N = C::N;
-    exists H = C::H;
-    exists II = C::II;
+    L := C::L;
+    N := C::N;
+    H := C::H;
+    II := C::II;
 }
 
 /// The module that interfaces with the outside world which always provides all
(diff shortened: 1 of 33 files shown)


rachitnigam:
@UnsignedByte this should let you encode the `Butterflies` stuff much more easily.

PR #368, opened 2023-10-18
Remove unused parameters generated for bundles

rachitnigam:
Follow up to #359 which make `idx` field in the IR an optional. We were still generating unused parameters because the AST would preemptively generate parameters for bundles when they were missing.


rachitnigam:
This is failing because even when the parameter is not used within the liveness specification of the interval of a bundle, it is still needed so that we can add assumptions about its length.
This makes me think that #359 doesn't actually make any sense.

rachitnigam:
#376 reverts the original change

PR #369, opened 2023-10-18
Model `let`-bound parameters in the IR

### LINKED ISSUE #362: Model `let` bindings in the IR
Currently, `let`-bindings for parameters are treated purely as syntactic sugar: they are eliminated in `from_ast` and not modeled in the IR. I think we should walk back this decision and model them explicitly in the IR. There are a couple of benefits from this:
* The error messages can reference information in a source-level format instead of the fully expanded version
* There is some hope that the solvers can find more syntactic similarities when solving constraints

The implementation should be straightfoward:
1. Model a new `ParamOwner` which tracks `let`-bound variables
2. Change `monomorphize` to track the values for `let`-bound variables

rachitnigam:
Fixes #362.


rachitnigam:
@UnsignedByte I've tried this out with #348 and it seems to slow down things. Regardless, you should experiment with these changes and see if they help at all. The benefit in error reporting might still make it worth merging.

rachitnigam:
`let`-bound values in the `with` section are still inlined with this PR. Maybe we should also turn them into `let` statements in the body of the component for symmetry and better error reporting.

rachitnigam:
The problem is that `build-domination` does not correctly reorder `let` statements in the IR causing monomorphize to fail.

UnsignedByte:
@rachitnigam Looks like build domination needs to do something like `Traversal` now...

The issue here is that instances can reference let parameters like so:
```rs
let P = 1;
Inst := new Comp[P];
```
But let parameters can also reference existential parameters
```rs
Inst := new Comp[1];
let P = Inst::P;
Inst2 := new Comp[P];
```
So we would have to check for dependencies (and fail on circular dependencies)...

UnsignedByte:
It might be better to instead do `build-domination` simply by moving instances, invokes, and plets to the top without changing their relative order (which makes sense because they can sort of be seen as definitions of the same hierarchy).

rachitnigam:
That wouldn’t work because an instance can use parameters which can use another instance defined parameter to compute its value.

UnsignedByte:
Yeah, that would mean the user has to define instances and plets in the right order. The alternate method of using topologicalsort is probably better, but also requires us to implement Hash for command I think.

rachitnigam:
I don’t know if we need hash. We can just track the dependencies between the commands right? Each instance defined parameter says which instance defined it.

UnsignedByte:
`TopologicalSort` Stores things in a hashmap - I guess I could instead create a temporary idx type?

UnsignedByte:
@rachitnigam LMK what you think!

rachitnigam:
LGTM!

PR #375, opened 2023-10-19
Bitvector encoding for proofs

### LINKED ISSUE #374: Explore bitvector encoding for parameters
We currently interpret all parameters as `Int` when encoding them in SMT. The benefit of this is that we get truly unbounded proofs over all integers. However, it is well known that integers are a lie and who wants to use a parameter that is greater than 2^64 anyways. The practical benefit of using a bitvector encoding is we might scale better when performing non-linear reasoning (like with integer division).

The big problem with switching to a bitvector encoding is handling overflow behavior. For example, the following program has a model:
```lisp
(declare-const x (_ BitVec 4))
(declare-const y (_ BitVec 4))
(declare-const z (_ BitVec 4))

(assert (= y (bvmul x z)))

(assert (and (bvugt x y) (bvugt z y)))
(check-sat)
(get-model)
```

```
(
  (define-fun x () (_ BitVec 4)
    #xa)
  (define-fun z () (_ BitVec 4)
    #x7)
  (define-fun y () (_ BitVec 4)
    #x6)
)
```

This is because the multiplication of `x` and `z` can overflow and produce a `y` smaller than both. The general trick to avoiding this problem is extending the bitwidth of the numbers and defining that they cannot take any values in a region that represents overflow. For multiplication, this would look like:
```
(declare-const x (_ BitVec 8))
(declare-const y (_ BitVec 8))
(declare-const z (_ BitVec 8))

(assert (= y (bvmul x z)))

; x and y don't overflow
(assert (and (bvuge x #b00000000) (bvult x #b00001111)))
(assert (and (bvuge y #b00000000) (bvult y #b00001111)))
(assert (and (bvuge z #b00000000) (bvult z #b00001111)))

(assert (and (bvugt x y) (bvugt z y)))
(check-sat)
```

We double the range of the bitvectors and say that they cannot take values bigger than 2^4. `z3` specifically also [provides custom operations](https://stackoverflow.com/questions/54611088/prove-overflow-checking-expression-is-right) to assert that a particular multiplication does not overflow.

Alive2 also uses [the bit-extesion encoding](https://github.com/AliveToolkit/alive2/blob/master/smt/expr.cpp#L824) to assert that particular computations do not overflow.

rachitnigam:
Fixes #374. This uses the encoding from the issue and instantiates both parameters and events as bitvectors and adds assertions to eliminate overflow.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -85,19 +85,6 @@ pub struct Opts {
     #[argh(option, long = "backend", default = "Backend::Verilog")]
     pub backend: Backend,
 
-    // Solver specific configuration
-    /// solver to use (default: cvc5): cvc5, z3
-    #[argh(option, long = "solver", default = "Solver::CVC5")]
-    pub solver: Solver,
-
-    /// solve assertions separately rather than all at once
-    #[argh(switch, long = "discharge-separate")]
-    pub discharge_separate: bool,
-
-    /// dump interactions with the solver in the given file
-    #[argh(option, long = "dump-solver-log")]
-    pub solver_replay_file: Option<String>,
-
     /// disable generation of slow FSMs in the backend
     #[argh(switch, long = "disable-slow-fsms")]
     pub disable_slow_fsms: bool,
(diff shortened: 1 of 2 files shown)


rachitnigam:
I'm going to merge this for now as a mechanism to quickly generate bounded proofs.

PR #377, opened 2023-10-19
Fixes a bug in compilation.

### LINKED ISSUE #378: Bug in guard synthesis
Not sure exactly how this is caused, but assigning ports to different invokes doesn't properly generate guards. A simple example that fails is in #377 and works as follows:
```rs
import "primitives/comb.fil";

/// Returns a const
comp ReturnConst<'G: 1>(
  in: ['G, 'G+1] 32
) -> (
  out: ['G, 'G+1] 32
) {
  out = in;
}

/// Test that checks multiple assignments to different invocations when all ports are active
comp main<'G: 2>(
  go: interface['G],
  in: ['G, 'G+1] 1
) -> (
  out1: ['G, 'G+1] 32,
  out2: ['G+1, 'G+2] 32
) {
  RC := new ReturnConst;

  c1 := new Const[32, 0]<'G, 'G+2>();
  c2 := new Const[32, 1]<'G, 'G+2>();

  rc1 := RC<'G>(c1.out);
  rc2 := RC<'G+1>(c2.out);

  out1 = rc1.out;
  out2 = rc2.out;
}
```

We expect an output of `out1 = 0` and `out2 = 1` but because of improper guard synthesis both `out1` and `out2` return `0`, as the invoke `RC<'G>(c1.out)` overrides the second one. This is because in guard synthesis destination ports are ignored when compiling connects, and it should be the opposite.

UnsignedByte:
Fixes #378


rachitnigam:
Does this also happen with `--disable-slow-fsms`?

UnsignedByte:
> Does this also happen with `--disable-slow-fsms`?

Pretty sure this was just a simple guard synthesis issue, the tests pass now so I'm auto-merging as its a tiny change.

rachitnigam:
Sounds good!

src/ir_passes/lower/build_ctx.rs
```diff
@@ -218,9 +218,9 @@ impl<'a> BuildCtx<'a> {
 
         log::debug!("Compiling connect: {}", self.comp.display(con));
 
-        // ignores the guard of the destination (bind check already verifies that it is available for at least as long as src)
-        let (dst, _) = self.compile_port(dst.port);
-        let (src, g) = self.compile_port(src.port);
+        // ignores the guard of the src (bind check already verifies that it is available for at least as long as dest)
+        let (dst, g) = self.compile_port(dst.port);
+        let (src, _) = self.compile_port(src.port);
```
rachitnigam, line 223: Pretty silly bug!

PR #379, opened 2023-10-20
Support multi-dimensional bundles

### LINKED ISSUE #351: Multidimensional Bundles
As the title says, we should have support for multidimensional bundles. I'm thinking something like this:
```
bundle a[10][20]: for<i, j> ['G, 'G + i + j + 1] 32;
```
where `i` ranges from `0` to `9` and `j` from `0` to `19`.

These can probably be eliminated by the IR if we group everything into one dimension like so:
```
bundle a[10*20]: for <k> ['G, 'G + k/10 + k-((k/10)*10) + 1] 32;
```

rachitnigam:
Fixes #351. Adds supports for multi-dimensional bundles. For now, just testing to make sure that the changes don't break any reasoning with single dimensional bundles. Once that works, I'll add support to the frontend for parsing multi-dimensional bundles.


rachitnigam:
`assign-check` is failing and disabling it makes `bundle-elim` fail. Need to debug.

rachitnigam:
Turned out to be a dumb issue!

rachitnigam:
There seems to be some compilation bug which causes the serializer and deserializer not to correctly generate their outputs. I think the post-monomorphization code is *probably* well-typed but this is the kind of thing that the old AST flow was good at giving a convincing guarantee about.

rachitnigam:
Haha, the merge fixed it! Seems like it is exactly the issue that @UnsignedByte fixed yesterday!

rachitnigam:
Okay, I was being to excited things still fail.

rachitnigam:
Okay, this is working now! @UnsignedByte I'd be interested in seeing if the pipelined FFT implementation becomes much easier to type check using this.

PR #381, opened 2023-10-20
Add assertion checking in monomorphize

UnsignedByte:
Checks assertions and assumptions in the monomorphize pass.


UnsignedByte:
@rachitnigam This is failing because of "generated" parameters that are used in implications, for example
`0 < _ < 5 => 0 < 2*_ < 10`. My plan is to just leave assertions not-fully-resolved with these parameters, and then resolving them in a later pass. Does this sound good?

rachitnigam:
Yeah, those assertions cannot be resolved even after monomorphization so we might as well ignore/remove them

rachitnigam:
There is a `relevant_vars` method on `PropIdx` now so you can use that to determine whether a particular proposition uses local parameters.

UnsignedByte:
> Yeah, those assertions cannot be resolved even after monomorphization so we might as well ignore/remove them

I agree as far as it's unreasonable to loop from 0 to 5 for the assertions.

UnsignedByte:
Hoist-facts is causing issues here because assertions are taken out of `if` statements, etc and use parameters that aren't defined yet. We should figure out a way to hoist facts for typechecking but retain the original assertions.

UnsignedByte:
This should be good to merge now because hoist-facts just got moved.

src/ir_passes/mono/monodeferred.rs
```diff
@@ -314,6 +326,45 @@ impl MonoDeferred<'_, '_> {
...
+                .base
+                .resolve_prop(self.monosig.base.get(prop).clone())
+                .get();
+
+            let reason = reason.ul();
+            let reason =
+                self.monosig.info(&self.underlying, self.pass, reason).get();
+
+            // After monomorphization, both assumes and asserts become asserts.
+            Some(ir::Fact::assert(prop, reason))
```
rachitnigam, line 364: We should a “generated” variant to ir::Reason which itself contains a Reason. Then, in diagnostics reporting, we can accumulate all the asserts that are generated from one source level assert and say which assert or assume caused the failure.

src/ir_passes/mono/utils/comp.rs
```diff
@@ -10,6 +10,9 @@ use super::{Base, IntoBase, Underlying};
 pub struct UnderlyingComp<'a>(&'a ir::Component);
 
 impl<'a> UnderlyingComp<'a> {
+    pub fn comp(&self) -> &'a ir::Component {
```
rachitnigam, line 13: We want to avoid exposing comp on underlying and instead expose methods that force the user of the struct to wrap things in Underlying.

UnsignedByte: The reason for doing this was because `relevant_vars` can't be run on an underlying component easily.

UnsignedByte: I guess I can rewrite it to use `IdxLike`?

rachitnigam: Yeah, probably!

rachitnigam: Or just define a wrapper method that unwraps the underlying comp and the underlying index

UnsignedByte:
@rachitnigam Is this what you meant by generated reasons, or should we intern them like expressions to avoid having to clone so many?

fil-ir/src/info.rs
```diff
@@ -815,6 +829,7 @@ impl Reason {
                     .with_message("event's delay must be greater than the instance's borrow length")
                     .with_labels(vec![live, ev])
             }
+            Reason::Generated { reason } => reason.diag(ctx),
```
rachitnigam, line 832: I would add some way for the label to say that this was generated after elaboration

rachitnigam:
Cloning reasons is fine for now IMO

PR #383, opened 2023-10-21
Implement instance borrow syntax

### LINKED ISSUE #109: Borrow instance syntax
Extend the instance creation syntax with option of defining the length of the "borrow" for the instance. This is meant to pre-commit to an interval during which the instance can be reused. I think that in general, parametric programs will need this in order for the shareability check (especially for #76)

rachitnigam:
Fixes #109. Adds one of two resource sharing checks using the new borrow syntax.


rachitnigam:
This enforces two new requirements:
1. That all used invocations use events in a way that they are within the range of the borrow's specified range.
2. The borrow's range has length less than or equal to the component's delay.

PR #386, opened 2023-10-25
Fix a bug in assumption generation for existential parameters

UnsignedByte:


rachitnigam:
The problem was that we did not add the path condition when generating the well-formedness constraint. This also changes the printer to print out the reason for an assertion being generated which makes debugging these kinds of problems easier.

PR #388, opened 2023-10-26
Set default solver to z3

UnsignedByte:
Changes the default solver used in discharge.


UnsignedByte:
Should be a quick merge, just sets the default solver to z3 and installs it for githb actions

PR #389, opened 2023-10-26
Move HoistFacts to own library

UnsignedByte:
Removes hoistfacts as a pass so that assumptions/assertions are not changed and destroyed in a component by hoistfacts/discharge.


src/ir_visitor/visitor.rs
```diff
@@ -287,19 +226,24 @@ where
...
+        opts: &'pass cmdline::Opts,
+        ctx: &'pass mut ir::Context,
+    ) -> Result<(), u64> {
         let mut visitor = Self::from(opts, ctx);
         for idx in ctx.comps.idx_iter() {
             visitor.clear_data();
-            visitor.visit((idx, opts, &mut *ctx).into());
+            let opts: &'comp cmdline::Opts = opts;
+            let ctx: &'comp mut ir::Context = &mut *ctx;
+            visitor.visit(Self::Data::new(idx, opts, ctx));
```
UnsignedByte, line 246: Is there any way to do what I'm trying to here? Been spending way too long on it and making no progress...

rachitnigam: The [lifetime bounds](https://doc.rust-lang.org/beta/reference/trait-bounds.html#lifetime-bounds) syntax is a bit confusing but your code currently says that `'pass` outlives `'comp` which is probably the opposite of what you want.

src/ir_visitor/visitor.rs
```diff
@@ -96,11 +49,13 @@ impl<T: Default> Construct for T {
...
 }
 
-/// A visitor for the commands by detaching them from the underlying component.
-pub trait Visitor
+/// A Visitor for the commands by detaching them from the underlying component.
+pub trait Visitor<'comp>
 where
     Self: Sized + Construct,
 {
+    type Data: VData<'comp>;
```
rachitnigam, line 57: Why does the data need the lifetime parameter to come from the trait instead of the struct that implements the trait? It seems we don't use this parameter except to pass to the VData field.

rachitnigam: I think you probably want an `ExtraData` field that stores other information that is beyond what `VisitorData` currently stores and expose that as a `self.get_data` and `self.get_mut_data` methods or something.

src/ir_visitor/visitor_data.rs
```diff
@@ -0,0 +1,62 @@
...
+        Self {
+            comp,
+            idx,
+            opts,
+            mut_ctx: ctx,
+        }
+    }
+}
+
+impl Drop for VisitorData<'_> {
```
rachitnigam, line 57: When you find yourself implementing drop, something has gone wrong :rofl:

UnsignedByte: Funnily enough that's the only thing that's been there since before this change - I added it way back when visitordata was first introduced.

src/ir_visitor/visitor.rs
UnsignedByte: It doesn't compile if you don't have `'comp` because the `new` function needs to generate a `Self` that lasts long enough. 
image
In the implementation of `new` I need `'comp` in the signature which also requires me to define it in `VData`.

UnsignedByte: Error if `VData` doesn't define `'comp`: 
image

UnsignedByte: I also can't just do the following because adding `'comp` changes the signature of the `new` function...
image

src/ir_visitor/visitor.rs
UnsignedByte: I actually want `'pass` to outlive `'comp` - `comp` should be alive for each component (each iteration of the loop) while `'pass` should be a reference to the context that stays alive for the entire pass.

src/ir_passes/discharge.rs
```diff
@@ -550,6 +547,12 @@ impl Visitor for Discharge {
     }
 
     fn start(&mut self, data: &mut VisitorData) -> Action {
+        self.to_prove = crate::utils::HoistFacts::hoist(&mut data.comp);
```
rachitnigam, line 550: style nit: Import `crate::utils::HoistFacts` at the top

rachitnigam (approved):
LGTM!

PR #391, opened 2023-10-27
Remove implications being generated with `false` left hand sides.

UnsignedByte:
The issue is caused by let-bound parameters eliding away into expressions like `0 == 1` in if statements, which results in assumptions/assertions generated inside to become `False => ??`.

The solution is to eliminate false branches and not convert them at all.

### DIFF
--- a/fil-ir/src/from_ast/astconv.rs
+++ b/fil-ir/src/from_ast/astconv.rs
@@ -986,9 +986,16 @@ impl<'prog> BuildCtx<'prog> {
             }
             ast::Command::If(ast::If { cond, then, alt }) => {
                 let cond = self.expr_cons(cond)?;
-                let then = self.commands(then)?;
-                let alt = self.commands(alt)?;
-                vec![ir::If { cond, then, alt }.into()]
+                if cond.is_false(self.comp()) {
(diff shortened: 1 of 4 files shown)


fil-ir/src/from_ast/astconv.rs
```diff
@@ -986,9 +986,16 @@ impl<'prog> BuildCtx<'prog> {
...
+                if cond.is_false(self.comp()) {
+                    // throw away the `then` branch to avoid false assumptions and assertions
+                    self.commands(alt)?
+                } else if cond.is_true(self.comp()) {
+                    self.commands(then)?
+                } else {
+                    let then = self.commands(then)?;
+                    let alt = self.commands(alt)?;
+                    vec![ir::If { cond, then, alt }.into()]
+                }
```
UnsignedByte, line 998: These things should really be pulled out of astconv soon :)

rachitnigam: Hm, this is problematic because if means that the typechecker will never see branches that are false and not report errors on them. If I write:
```
if true {
  // no bugs
} else {
  // bugs here
}
```
The type checker won't report any errors...

PR #393, opened 2023-10-30
Use docker images in testing

UnsignedByte:
Tries using dockerfile for tests. ~~Currently uses `latest` because we don't have tagged versions.~~

This splits the `test` action into `3` steps:
1. Use `git log` to get the hash of the last commit that edited `Dockerfile`, and also check if that ghcr tag already exists on the remote.
2. Conditionally run a build-push action if it doesn't exist, which just builds and pushes the docker image of that specific commit with the right tag.
3. Run test using that docker image.


UnsignedByte:
The test action takes only 5-6 mins now (compared to 15) because it doesn't have to rebuild all dependencies which is cool.

UnsignedByte:
This should be good to merge but I want to pass this by you first because we are pulling the `latest` docker image which I don't think will cause issues as filament is rebuilt but I'm not sure.

rachitnigam:
Are we building latest on every commit to main? If so, there is a problem: depending on when a PR is opened, it might be tested with the wrong latest (if the previous commit to main hasn’t completed the build yet).

We could do something to rebuild the docker image for PRs that touch the dockerfile and otherwise use latest but that requires some GH actions hacking.

UnsignedByte:
Yeah, this solution would definitely break on branches that mess with the dockerfile and need those changes to pass tests, which was why I wanted to pass it by you first.

rachitnigam:
One dumb option is using GH actions cache and index it using a hash of the Dockerfile. Then, the run would only trigger if the cache doesn't contain an image for a particular hash.

UnsignedByte:
Is that a dumb option? My thought was to just use git fetch to check whether the dockerfile had been modified but this sounds better honestly...

UnsignedByte:
@rachitnigam Probably a severely overengineered solution but this should work. I've updated the PR description to describe what it does.

rachitnigam:
@sampsyo might enjoy the hackery in this! LGTM!

PR #396, opened 2023-11-03
Adds flopoco to the floating point tests

UnsignedByte:
Plays around with some gen framework stuff


rachitnigam:
Did this end up not working?

PR #398, opened 2023-11-03
Flopoco playground

UnsignedByte:
Uses a very hacky and non-foolproof solution that allows for multiple instantiations of a gen framework - obfuscating all non-top modules.

A better solution would be to write a pass that gathers all externs and uses `verible-verilog-syntax` to find their module names, and assigns unique random names to all non-exposed modules (helper modules).


UnsignedByte:
Closing this as it has since been replaced by #432.

PR #400, opened 2023-11-05
Add Bazel and XLS to docker image

gabizon103:
Adds Bazel and XLS to the docker image. XLS is quite large so it takes some time to build the first time. Was having issues building Bazel on my Mac, but it works fine on Havarti.


rachitnigam:
@gabizon103 looks like the docker image is failing to find the bazel package in `apt`. We can debug this later since this is not a blocker for experiments right now.

rachitnigam:
@UnsignedByte mentioned that this has been merged into #398

PR #403, opened 2023-11-17
Fix typos in documentation

Xeyler:
I've been fixing typos as I read the documentation.

I'm taking my first class on hardware design this semester. I'm very new to it. It hasn't been too difficult understanding hardware concurrency, but I don't understand at all how computer engineers pipeline modules and match the timings and delays between them. If an engineer were to work with a large number of modules, it would become very difficult to keep track of signal validity between them.

I really like the solution that filament provides in response to the issue! This is a really cool piece of software.


rachitnigam:
Thank you for contributing back! If you end up playing with Filament, we'd love to hear about your experience!

PR #407, opened 2023-12-16
XLS Docker Image

UnsignedByte:


rachitnigam:
Looks like XLS is only supported on x86? If so, let's out XLS into its own Dockerfile and keep the Filament one more minimal.

UnsignedByte:
How do we plan on testing the XLS gen framework in GHA then? A separate XLS install?

rachitnigam:
GHA only needs x86 images i think?

Issue #411, opened 2024-02-15
BLAS evaluation


gabizon103:
Taking inspiration from [this paper](https://spcl.inf.ethz.ch/Publications/.pdf/fblas.pdf), implement some level 1 and 2 BLAS kernels and parameterize them similarly. Specifically:
- a `GEMV` implementation that is parameterized over matrix tiling and hardware reuse
- a `DOT` implementation that is parameterized over hardware reuse
- a `SCAL` implementation that is parameterized over hardware reuse (maybe redundant?)
- an `AXPY` implementation that is parameterized over hardware reuse

If time permitting, also show that modules can be chained together using output parameters. The linked paper uses FIFOs to make some computations blocking in their compositions, but since this isn't possible in Filament we can just use output parameters instead. 

These implementations seem straightforward, so that we have an easy way to show that Filament can be used for design space exploration.

rachitnigam:
One note about #413: A cool thing enabled by being parameteric over the multiplier's (and adder's) timing behavior is we can implement precision optimizations: we can support both floating-point and fixed-point implementations of all the kernels.

Showing off this example might be useful to demonstrate how when doing DSE, there are many different kinds of "correctness" you need to care about and Filament doesn't track all of them (in this case,  bit-precision in computations).

PR #412, opened 2024-02-18
Add dot-product kernel with multiplier reuse

rachitnigam:
Starting work on #411 with a simple dot-product kernel that expresses multiplier reuse! Tagging @gabizon103 for review!


rachitnigam:
Okay, this implementation works and has a test now! We should probably generalize the test a bit more and make we can execute `Dot` multiple times.

rachitnigam:
@gabizon103 can you take over the merging of this PR. Either:
1. Decide that your implementation strictly subsumes this and therefore this PR should be closed, or
2. Add this in a different folder as a comparison point

gabizon103:
Yes, will take care of it!

PR #413, opened 2024-02-19
BLAS dot product kernel

gabizon103:
Implements a dot product kernel that is parameterized over multiplier reuse, from #411


apps/blas/dot/dot.fil
```diff
@@ -0,0 +1,217 @@
...
+      m := M<'G+mul_start>(x{j*N..(j+1)*N}, y{j*N..(j+1)*N});
+      mult_out{j}{0..N} = m.out{0..N};
+    // any other use should use registered inputs
+    } else {
+      x_concat := new ConcatBundle[W, N]<'G>(x{j*N..(j+1)*N});
+      y_concat := new ConcatBundle[W, N]<'G>(y{j*N..(j+1)*N});
+      x_concat_reg := new Shift[W*N, mul_start]<'G>(x_concat.out);
+      y_concat_reg := new Shift[W*N, mul_start]<'G>(y_concat.out);
+      x_bundle_reg := new SplitWire[W, N]<'G+mul_start>(x_concat_reg.out);
+      y_bundle_reg := new SplitWire[W, N]<'G+mul_start>(y_concat_reg.out);
```
rachitnigam, line 122: This pattern of concat -> shift -> split seems reusable. We could consider changing the shift register definition to accept bundles instead of values and default it to 1

apps/blas/dot/dot.fil
```diff
@@ -0,0 +1,217 @@
...
+) with {
+  some L where L > 0;
+  let reuses = S / N;
+} where W > 0,
+        L > 0,
+        S % N == 0
+{
+
+  M := new Multipliers[W, N];
+  RA := new RAdd[W, N];
```
rachitnigam, line 90: Use the instance borrow syntax to specify how long both of these components are reused for.

apps/blas/dot/dot.fil
```diff
@@ -0,0 +1,217 @@
...
+  let S = 16;
+  let W = 32;
+  let II = S/N;
+  some L where L > 0;
+} where L > 0 {
+
+  Dx := new Dot[W, S, N];
+
+  bundle x[S]: for<k> ['G, 'G+1] W;
+  bundle y[S]: for<k> ['G, 'G+1] W;
```
rachitnigam, line 53: “for” is not needed because the interval doesn’t use k

apps/blas/dot/dot.fil
gabizon103: Wouldn't the time over which I reuse them depend on the output parameters of the instances? For the multiplier specifically, I would imagine that the interval would look like `['G, 'G + reuses*M::II + M::L]`, which we can't express.

apps/blas/dot/dot.fil
```diff
@@ -0,0 +1,217 @@
...
+        L > 0,
+        S % N == 0
+{
+
+  M := new Multipliers[W, N];
+  RA := new RAdd[W, N];
+
+  // needed for typechecking
+  assume RA::L == log2(N);
+  assume M::II <= reuses;
```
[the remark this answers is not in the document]
gabizon103: I agree it is sketchy because it wouldn't always be safe. But otherwise, the compiler throws errors that the invocation of `M` triggers too often since we want `Dot`'s delay to be `reuses`

apps/blas/dot/dot.fil
rachitnigam: Why can't we express that? Can't we write:
```
M := new Multipliers[W, N] in ['G, 'G + reuses*M::II + M::L];
```

gabizon103: image

This is something that should be supported, right? As in we should be able to figure out what `M:II` and `M::L` are just from the instantiation, and then use them in the borrow interval.

apps/blas/dot/dot.fil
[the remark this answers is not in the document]
gabizon103: Totally makes sense. What should I do here, then? Because if we abstract over multiplier implementation, we need some way to reason about the multiplier's II in comparison to the entire module's II, and I'm not sure how else to do that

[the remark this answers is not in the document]
gabizon103: If you look again, I was able to get rid of this `assume`; I was using `<'G: II>` in the  signature of `Multipliers`. Now I changed that just to `<G: 1>` but kept `II` as an output parameter (so it can be accessed elsewhere)

apps/blas/dot/dot.fil
gabizon103: #417

PR #414, opened 2024-02-22
Slang Module Name management for filament

UnsignedByte:
Use Slang to rename modules deterministically rather than using the hacky obfuscation tactic


rachitnigam:
Currently has a lot of generated files so I'm assuming not ready for review

rachitnigam:
We could consider using Morty: https://github.com/pulp-platform/morty

PR #417, opened 2024-02-22
Add shift register interface that accepts bundles

gabizon103:
Made a new module with bundle interface. Uses existing `Shift` module.

### DIFF
--- a/primitives/state.fil
+++ b/primitives/state.fil
@@ -1,3 +1,5 @@
+import "primitives/reshape.fil";
+
 extern "state.sv" {
    /// ANCHOR: register
    // A register that can extend the lifetime of a signal to any required length.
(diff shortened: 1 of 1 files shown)


rachitnigam:
Why can't we just extend the `Shift` module and make `S=1`? I would also rename the parameters so that `N` is the number of inputs and `D` is the depth of the shift register

rachitnigam:
Nice! As we start building this higher-level abstractions, we should also think of some optimizations to do at the Filament level. For example, `ConcatWire` and `SplitWire` definitions can be inlined into `Shift` for basically no cost and for the `N=1` case, eliminate a bunch of wires. Those wires are probably not that expensive to begin with but still worth thinking about building up optimization infrastructure.

PR #421, opened 2024-02-28
Fix shift reg indices

gabizon103:
Made a silly mistake that I didn't catch in the other PR, didn't realize you need to do `a{0..N} = b{0..N}` instead of just `a = b` when doing bundle assignments.


rachitnigam:
Oh man, annoying bug here! We should open a bug-report about this and work towards improving the experience with bundle accesses in general.

PR #424, opened 2024-03-02
`axpy` BLAS implementation

gabizon103:
This is parameterized over the number of multipliers and adders used. Might be a bit of overkill, but better to have it than not, especially if we want to use floating point. The multiplications are additions are "fused" where possible (depends on the ratio of multipliers to adders). @sampsyo also mentioned that in practice it might be better to have the number of multipliers and adders be the same, in the case where a synthesis tool can infer some kind of fused DSP multiply-add.


apps/blas/axpy/axpy.fil
```diff
@@ -0,0 +1,41 @@
...
+        N > 0,
+        M > 0,
+        A > 0
+{
+  // Perform a*x
+  Sc := new Scal[W, N, M];
+  s := Sc<'G>(x{0..N}, a);
+
+  // register y
+  y_reg := new Shift[W, Sc::L, N]<'G>(y{0..N});
```
rachitnigam, line 30: Interesting! There is a different version of this that "fuses" the multiplication and adding so that we don't need to wait for all of `a*x` to be computed before computting `a*x+y`. Thoughts on how hard that would be express? I almost want to say that `Scal` (I'm assuming scalar product) should be implemented in terms of axpy.

gabizon103: Ooh, interesting! I will try that, it sounds like it shouldn't be too hard to express!

apps/blas/axpy/test.expect
```diff
@@ -0,0 +1,2 @@
+{"out_0": {"0": [0], "1": [0]}, "out_1": {"0": [1], "1": [2]}, "out_2": {"0": [2], "1": [4]}, "out_3": {"0": [3], "1": [6]}, "out_4": {"0": [4], "1": [8]}, "out_5": {"0": [5], "1": [10]}, "out_6": {"0": [6], "1": [12]}, "out_7": {"0": [7], "1": [14]}, "out_8": {"0": [8], "1": [16]}, "out_9": {"0": [9], "1": [18]}, "out_10": {"0": [10], "1": [20]}, "out_11": {"0": [11], "1": [22]}, "out_12": {"0": [12], "1": [24]}, "out_13": {"0": [13], "1": [26]}, "out_14": {"0": [14], "1": [28]}, "out_15": {"0": [15], "1": [30]}, "cycles": 8}
```
rachitnigam, line 1: This IO format is getting annoying to read with these matrix programs 😅. We should change the test harness so it can accept inputs in other formats too!

gabizon103: yeah big agree on this lol

rachitnigam:
@gabizon103 is there anything blocking this merge?

gabizon103:
Nope!

PR #435, opened 2024-03-18
Using Hypermapper 

UnsignedByte:
Messing around with hypermapper and bayesian optimization.

The eventual goal for this is first to write all the necessary features to make last year's evaluation significantly easier.

Ideally, we would be able to provide some form of interface json that says the following:
```json
{
  "inputs": {
      "some parameter": [
          "range": [0, 10],
          "type": "int"
      ],
      ...
  },
  "targets": {
     "registers", "luts", "throughput"
  }
}
```
or something equivalent, and hypermapper would automatically tweak the necessary values for us.

Eventually, with possibly the merge of #441  and something similar, we could have filament generate the input parameters and their ranges automatically given a flag like `--dump_interface`, and this entire process might be automateable.


rachitnigam:
Looks cool! What is the game plan with this?

UnsignedByte:
Closing this as it has been moved to [parafil-evaluation](https://github.com/cucapra/parafil-evaluation).

PR #438, opened 2024-04-09
rot kernel

gabizon103:


apps/blas/rot/rot-alt.fil
```diff
@@ -0,0 +1,131 @@
...
+  out_1[N]: ['G+L, 'G+L+1] W,
+  out_2[N]: ['G+L, 'G+L+1] W
+) with {
+  some L where L > 0;
+  some II where II > 0;
+} where W > 0,
+        N > 0,
+        M > 0,
+        A > 0,
+        M % 4 == 0,   // need to do at least 4 multiplies at once
```
rachitnigam, line 27: Tangential fly-by comment:  maybe we should add some syntax to allow asserts to add a reason for the assertion so we can generate better errors. For example when one of these asserts trigger, the comment could show up as the rationale for the restriction.

PR #439, opened 2024-04-11
Gen config

UnsignedByte:
Adds the ability to configure compilation parameters through a config file.

The user provides a toml file as follows:
```
[globals.flopoco]
conf = "target=Virtex6 frequency=200"
```
to override the `globals.conf` element of `def.fil`.

The user can also provide parameters to the `main` component by doing the following:
```
params.main = [8, 32]
```


tests/errors/binding/main-concrete.expect
```diff
@@ -0,0 +1,11 @@
+---CODE---
+1
+---STDERR---
+error: Default values for parameters in the main component must be concrete
+  ┌─ tests/errors/binding/main-default.fil:2:19
+  │
+2 │ comp main[?A=31, ?B=1+W]<'G:1>() -> () {
```
rachitnigam, line 7: The parameter `W` is not bound in this interface?

tests/errors/binding/main-param.expect
```diff
@@ -0,0 +1,11 @@
+---CODE---
+1
+---STDERR---
+error: `main' requires at least 1 parameters but 0 were provided
+  ┌─ tests/errors/binding/main-param.fil:2:6
+  │
+2 │ comp main[A]<'G:1>() -> () {
+  │      ^^^^ `main' requires at least 1 parameters but 0 were provided
```
rachitnigam, line 8: Can we improve the error message for this case and provide some extra guidance on how to parameterize `main` because that is different from other components

tests/gen/fp-add/add_200.expect
```diff
@@ -0,0 +1,12 @@
+{
```
rachitnigam, line 1: What are these new tests for?

tests/gen/shift/6.toml
```diff
@@ -0,0 +1 @@
+params = [32, 6]
```
rachitnigam, line 1: Can we use a slightly more explicit syntax here:
```
main.params = [32, 6]
```

In the future, we might want multiple externally-visible components.

rachitnigam:
Some suggestions to improve the error messages and question about new tests. Also, tests should ideally go in the `tests/` folder.

tests/errors/binding/main-concrete.expect
UnsignedByte: Looks like this wasn't caught because it was superseded by the error about concreteness. Fixed!

tests/gen/fp-add/add_200.expect
UnsignedByte: These are to check that changing the gen config parameters actually does change the generated verilog; the tests check that the latency of the module changes with different frequencies etc.

tests/gen/shift/6.toml
UnsignedByte: Sure, but just as a note if we want this to be supported for multiple components we could no longer have the whole `[A, B]` param system and would have to build something different anyway.

UnsignedByte:
> Some suggestions to improve the error messages and question about new tests. Also, tests should ideally go in the `tests/` folder.

As far as I can tell these are all in `tests/`? I created the `gen` folder because the tests have a special way of being invoked that is different from others, should I change that?

tests/gen/fp-add/add_200.expect
rachitnigam: Got it!

tests/gen/shift/6.toml
rachitnigam: What do you mean?

tests/errors/binding/main-param.expect
rachitnigam: This comment was resolved but not actually addressed?

tests/gen/shift/6.toml
UnsignedByte: Because normal parameters run the risk of being double bound by both the invoking component and the configuration, so we would want to be able to instead specify parameters that are only modified by the configuration file (like in #441).

tests/errors/binding/main-param.expect
UnsignedByte: I added some more info to the error message in the code? Do you mean to add more than this?

rachitnigam: I don't see the change. I meant we should say something like: "`main` is a special external component. Parameterizing it correctly requires using the the parameter file/cmdline arguments"

tests/errors/binding/main-param.expect
```diff
@@ -0,0 +1,14 @@
...
+
+error: Incorrect parameter bindings provided to top-level component main
+ = Parameter bindings should be provided via the `--bindings` flag in a `.toml` file.
```
UnsignedByte, line 11: It's here (not sure why it wasn't showing before) @rachitnigam

PR #442, opened 2024-05-17
Use full path for filament target in fud configuration

UnsignedByte:
Current setup means fud will only run in the filament home directory.


rachitnigam:
LGTM!

PR #443, opened 2024-05-28
Syr kernel

gabizon103:
Implements syr, which computes `a*x*x^T + A`, where `a` is a constant, `x` is a vector, and `A` is a symmetric matrix.


rachitnigam:
Awesome stuff @gabizon103!!

Issue #299, opened 2023-09-01
Attributes for components


rachitnigam:
Going forward, it would be useful to expose some mechanism of adding attributes to components (and other things like instances) to control code generation. For example, we can put an attribute to force generation of an FSM with a particular delay:

```
#[slow_fsm=5]
comp main<G:10>(...) {...}
```

The attribute can then be used by the lowering pass to generate the FSM of the specified delay (and circumvent the need for #273 for some time).

UnsignedByte:
A quick question on attributes before we implement them should be about scoping of different attributes. For example, say I have a two-event component as follows:
```rust
#[slow_fsm=5]
comp main<G: 10, L: 9>(...) -> (...) {...}
```
which is currently not super realistic to create currently but nonetheless exists, and could become important in dynamic filament. A component level attribute like `slow_fsm` would not be able to distinguish between the two events and which one to apply the slow FSM to.

Even further we may have attributes we want at the port level, such as information about how bundles could be lowered at the top level. The simplest example of this is with the spiral DFTs, where the verilog emitted can have say, `4` input ports for a `16` point FFT where the first input represents the `0th, 4th, 8th, 12th` input in each cycle. We have discussed before possibly having some form of lowering where a bundle like
```
A[4]: for<i> ['G+i*4, 'G+i*4+1] 32
```
would be lowered into a single port that is reused at different times for different behaviors. This behavior could be decided via attribute, where `#[collapse_bundle]` or something similar could decide whether a bundle is collapsed into one port or split into multiple.

UnsignedByte:
Of course component level attributes could come first but I do think this is an idea that might be worth exploring in the future.

rachitnigam:
I think the approach you've taken---to implement the simplest thing first (component-level attributes)---is probably the right one. As we add more features to the language (real modules, real scoping rules), we can investigate more fine-grained attributes.

PR #449, opened 2024-08-28
Add attributes for components and implement `toplevel` and `counter_fsm` attributes

UnsignedByte:
Adds the language capability for component level attributes, implementing #299.

Currently adds one attribute, `slow_fsms=n` telling the compilation pass to use slow counter based FSMs with `N` states per count. Requires that `n` divides the II of the component.

This currently has a few implementation details and limitations that need to be discussed:
* This is only available at the component level, so this does not allow us to generate different types of FSMs for different events.
* The delay of a slow FSM must divide the II of the event (in multi event components this means it must divide the GCD), there is not a really nice way to detect this sort of issue.
* `0` is currently used to specify simple FSMs.
* The attribute is now added automatically by a pass rather than decided at compile time, which subscribes it to the  same limitations as everything else.

As a proof of concept, this also adds the `#[counter_fsm]` attribute, which allows the user to specify whether an attribute instantiates a counter fsm or not.


crates/filament/src/ir_passes/fsm_attributes.rs
```diff
@@ -0,0 +1,48 @@
+use crate::ir_visitor::{Visitor, VisitorData};
+use fil_ast as ast;
+use fil_ir::TimeSub;
+
+/// Sets the proper FSM Attributes for every component
+#[derive(Default)]
+pub struct FSMAttributes;
+
+impl FSMAttributes {}
```
rachitnigam, line 9: Empty `impl`

crates/ast/src/attribute.rs
```diff
@@ -0,0 +1,78 @@
...
+pub enum Attr {
+    Bool(BoolAttr),
+    Num(NumAttr, u64),
+}
+
+/// A set of attributes attached to a component
+#[derive(Default, Clone)]
+pub struct Attributes {
+    bool_attrs: BitFlags<BoolAttr>,
+    num_attrs: EnumMap<NumAttr, Option<u64>>,
```
rachitnigam, line 33: Is there a reason this can be `Option<u64>`? Wouldn't that make it a `BoolAttr`?

crates/ast/src/attribute.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    /// Get the value of a numeric attribute
+    pub fn get_numeric(&self, attr: NumAttr) -> Option<u64> {
+        self.num_attrs[attr]
+    }
+
+    /// Set the value of a numeric attribute
+    pub fn set_numeric(&mut self, attr: NumAttr, value: Option<u64>) {
+        self.num_attrs[attr] = value;
+    }
```
rachitnigam, line 77: We probably want a unified `get` method which returns `1` for `BoolAttr` and the numeric value for others.

tests/errors/well-formed/invalid-attributes.expect
```diff
@@ -0,0 +1,6 @@
+---CODE---
+101
+---STDERR---
+thread 'main' panicked at crates/ast/src/parser.rs:828:96:
+Found unknown attribute flag "invalid_attribute"
```
rachitnigam, line 5: Can we make the the compiler generate a nice error message like the Calyx frontend does: https://github.com/calyxir/calyx/blob/main/calyx-frontend/src/parser.rs#L282

crates/ast/src/attribute.rs
UnsignedByte: Would this not limit numeric attributes to 0 as a default value? I was thinking of adding passes that set the values if they dont exist (see FSMAttributes pass)

rachitnigam: Hmm, I was thinking that if the `Attr` was not defined, it would not be present in the map (this is how Calyx attributes work).

UnsignedByte: I think the map is represented as an array under the hood so everything is initialized to default (0)...

UnsignedByte: I chose to use this instead of the calyx linkedhashmap due to performance, but I could do it that way instead?

rachitnigam: Okay, got it. `Option<T>` is bigger than just `T` which is why I was wondering. Either way, doesn't matter too much I suspect.

crates/filament/src/ir_passes/lower/build_ctx.rs
```diff
@@ -237,7 +234,17 @@ impl<'a> BuildCtx<'a> {
                 );
             };
             let delay = delay.concrete(self.comp);
-            let typ = FsmType::new(states, delay, self.disable_slow_fsms);
+            let typ = match self.comp.attrs.get_numeric(ast::NumAttr::SlowFSM) {
+                None => FsmType::Simple(states),
+                Some(0) => FsmType::Simple(states), // 0 is a special case for simple FSMs
+                Some(n) => {
+                    assert!(
```
rachitnigam, line 241: Can we move this into the well-formedness check earlier in the compiler pipeline?

rachitnigam (approved):
LGTM! Thanks for getting us started on attributes. I think the `toplevel` attribute is also a good start and maybe we want to define something like a `no_mangle` attribute to preserve the names of several Filament components in the generated Verilog.

Minor requests for changes. Merge at your discretion after fixing.

crates/filament/src/ir_passes/lower/build_ctx.rs
UnsignedByte: Ok, will do but I will also keep this assert just in case things go wrong somewhere in the compiler.

rachitnigam: Ah yes, this is the correct choice!

rachitnigam: Sure, that also seems reasonable!

UnsignedByte:
Ok continuing to think about this has made the problem more and more annoying...

Say we have a boolean flag `dummy`. Dummy affects the way the compiler works, and can be enabled or disabled (with performance differences). There are three ways we may want to communicate dummy to the compiler:

1. `#[dummy]` would say "please enable dummy for this comp"
2. `#[no(dummy)]` would say "please do not enable dummy"
3. Nothing at all may instead tell the compiler "please decide whether to do dummy at your discretion.

Therefore, to represent any given attribute we would want an `Option<bool>` (or `Option<u64>`. This seems to be a somewhat reasonable solution for now, but as this is going down rust's path it may be prudent to think a little more about this before we go even further into preprocessor logic.

crates/filament/src/cmdline.rs
```diff
@@ -104,9 +100,9 @@ pub struct Opts {
     /// backend to use (default: verilog): calyx, verilog
     #[argh(option, long = "backend", default = "Backend::Verilog")]
     pub backend: Backend,
-    /// disable generation of slow FSMs in the backend
-    #[argh(switch, long = "disable-slow-fsms")]
-    pub disable_slow_fsms: bool,
+    /// disable generation of counter FSMs in the backend
```
rachitnigam, line 103: What is a counter FSM? Or rather, how is a new user supposed to know what a counter FSM is?

rachitnigam (approved):
LGTM! Generally I would prefer to separate out the PR into two pieces since some of the code changes add the AST visitor back and the other parts add the attribute infrastructure. Up to you if you want to merge it as one but I would update the PR name in that case.

UnsignedByte:
Closing this because we are splitting it instead into two separate PRs.

PR #458, opened 2024-09-11
Add AST Passes and the #[toplevel] attribute.

UnsignedByte:
Adds a component attribute allowing the user to specify a toplevel component other than `main`. If not specified, `main` is retained as the default toplevel component. To do this, it re-introduces AST Passes through an AST Visitor, and implements the same macros as ir passes for generating pass pipelines.

Errors are thrown if:
* Multiple toplevel components are specified
* An extern is specified as a toplevel component

### DIFF
--- a/crates/ast/src/component.rs
+++ b/crates/ast/src/component.rs
@@ -1,3 +1,5 @@
+use crate::BoolAttr;
+
 use super::{Command, Id, Signature};
 use fil_gen as gen;
 use gen::GenConfig;
@@ -44,6 +46,7 @@ impl Component {
     }
 }
 
+#[derive(Default)]
 pub struct Namespace {
     /// Imported files
     pub imports: Vec<String>,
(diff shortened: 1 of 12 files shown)


crates/ast/src/component.rs
```diff
@@ -103,8 +94,14 @@ impl Namespace {
     /// Get the index to the top-level component.
     /// Currently, this is the distinguished "main" component
     pub fn main_idx(&self) -> Option<usize> {
-        self.components.iter().position(|c| {
-            c.sig.name.inner() == &Id::from(self.toplevel.clone())
-        })
+        self.components
+            .iter()
+            .position(|c| c.sig.attributes.get(BoolAttr::TopLevel) == Some(1))
```
rachitnigam, line 99: We should define a "has" method for these kinds of tests

rachitnigam (approved):
LGTM!

PR #459, opened 2024-09-11
Set up .gitattributes to change line endings to `LF`

UnsignedByte:
Because `runt` uses `LF` instead of `CRLF` line endings, certain tests that for some reason have `CRLF` endings are failing. this should fix the issue.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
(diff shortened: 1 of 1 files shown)


rachitnigam:
Awesome! We should also fix runt to do the right thing.

PR #460, opened 2024-09-16
Patch for Spade -> Filament

ethanuppal:
NOT READY FOR REVIEW

### DIFF
--- a/crates/filament/src/resolver.rs
+++ b/crates/filament/src/resolver.rs
@@ -32,7 +32,7 @@ impl From<&cmdline::Opts> for Resolver {
 }
 
 impl Resolver {
-    fn new(lib: Vec<PathBuf>, input: PathBuf) -> Self {
+    pub fn new(lib: Vec<PathBuf>, input: PathBuf) -> Self {
         let mut lib = lib.clone();
         lib.push(From::from("."));
         Self {
--- a/crates/filament/src/resolver.rs
+++ b/crates/filament/src/resolver.rs
@@ -1,8 +1,8 @@
-use crate::cmdline;
+use crate::{cmdline, ir_visitor::Construct};
 use fil_ast as ast;
 use fil_utils::{Error, FilamentResult};
 use std::{
-    collections::HashSet,
+    collections::{HashMap, HashSet},
     fs,
     path::{Path, PathBuf},
 };
@@ -22,7 +22,7 @@ impl From<&cmdline::Opts> for Resolver {
         let mut lib = opts.library.clone();
         // Do this to resolve the issue mentioned in `cmdline.rs`
         // where default values don't work with argh
-        lib.push(PathBuf::from("."));
+        lib.push(From::from("."));
         Self {
             lib,
             input: opts.input.clone(),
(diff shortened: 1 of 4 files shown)


primitives/core.sv
```diff
@@ -0,0 +1,8 @@
+module Wire #(
+  parameter WIDTH = 32,
+) (
+  input wire logic [WIDTH-1:0] left,
+  output wire logic [WIDTH-1:0] out
+);
+  assign out = left + right;
```
rachitnigam, line 7: What's going on here? `right` is not defined...

crates/filament/src/resolver.rs
```diff
@@ -32,6 +32,16 @@ impl From<&cmdline::Opts> for Resolver {
 }
 
 impl Resolver {
+    pub fn new(lib: Vec<PathBuf>, input: PathBuf) -> Self {
```
rachitnigam, line 35: What's the use case for this method?

primitives/core.fil
```diff
@@ -1,2 +1,10 @@
 import "./comb.fil";
-import "./state.fil";
\ No newline at end of file
+import "./state.fil";
+
+extern "core.sv" {
+   comp Wire[WIDTH]<'G, 'L>(
```
gabizon103, line 5: What would this be for? I imagine it would be for renaming ports, but it can probably be done with Filament bundles.

If you have some port `p0`, I guess you would want to pass it to some instantiation `add_in := new Wire[W]<'G>(p0)` to get a port named `add_in.out` that would be passed into an adder?

If you have `p0`, you can define a bundle `bundle add_in[1]: 'G, 'G+1] W` and then make the assignment `add_in{0} = p0`. Then you'd use `add_in{0}` wherever. To me, this is also more readable.

primitives/core.sv
ethanuppal: I'm still making changes, so this is a draft PR and not ready for review. There's a lot of broken stuff right now.

crates/filament/src/resolver.rs
ethanuppal: I'm still making changes, so this is a draft PR and not ready for review. There's a lot of broken stuff right now.

ethanuppal: I want to make a resolver without constructing an instance of the filament command line options.

primitives/core.fil
ethanuppal: It makes implementing Spade's `Alias` operation easier.

rachitnigam: I think we should move Spade-specific stuff to a `spade.fil` file instead instead of adding things to the standard library.

primitives/core.sv
rachitnigam: I think you might have requested a review. If that was unintentional, then tag me when ready for review.

ethanuppal: It might have automatically done that

primitives/core.fil
ethanuppal: Good idea, I was thinking that too

ethanuppal:
Ok don't actually think I need to expose `SigMap`? but I'll keep it in for now in case I do

PR #465, opened 2024-10-08
Fix entrypoint setting with calyx

UnsignedByte:
Turns out setting the `toplevel` attribute is not enough for calyx, and we need to explicitly set an entrypoint.


rachitnigam:
LGTM!

PR #471, opened 2024-10-18
Support running discharge after monomorphization and before lowering.

### LINKED ISSUE #469: Discharge fails if any indices have been invalidated.
Discharge works with `DenseIndexInfo`s, meaning if any parameters/events/ports have been invalidated before discharge this throws an error. The culprit is here:
https://github.com/cucapra/filament/blob/2fe2b317fc96ed6453dada529c08289ad08ec4f6/crates/filament/src/ir_passes/discharge.rs#L591-L608

This causes issues because `.ports` and `.events` automatically only iterate over valid indices. We can solve this in two ways:
*  Some sort of "cleanup" pass that removes invalid indices (likely having to reconstruct the entire IR)
* Change `.push` to extend the `DenseIndexInfo` by default. This does not work in the case of discharge though, where `SExpr` is not `Default`.

UnsignedByte:
- Fixes #469 
- Adds a second type checking interval checking etc pass after monomorphization, to ensure that checks still pass afterwards. Also reworks a few passes to make sure that Discharge can still run.

### DIFF
--- a/apps/fft/test/.gitignore
+++ b/apps/fft/test/.gitignore
@@ -1,0 +1,1 @@
+rand.json


rachitnigam:
Also fixes the problem with Exprs you mentioned? Also, LGTM

UnsignedByte:
Fixes things now. I added discharge + typechecking after monomorphize if that seems reasonable?

crates/ir/src/fact.rs
```diff
@@ -357,6 +359,19 @@ impl PropIdx {
             }
         }
     }
+
+    pub fn valid(
+        &self,
+        ctx: &(impl Ctx<Prop>
```
rachitnigam, line 365: Use a context bound: `valid<Ctx>(..) where Ctx: ...`

rachitnigam:
Minor comment, LGTM!

crates/ir/src/fact.rs
UnsignedByte: This doesn't match with the existing functions `relevant_vars` etc though. Should I change all of them or just valid?

```
pub fn relevant_vars(
        &self,
        ctx: &(impl Ctx<Time> + Ctx<Prop> + Ctx<Expr>),
 ) -> ...
 ```

PR #474, opened 2024-10-28
Reworking attributes to match Ctx

UnsignedByte:
This reworks attributes to use a system like the `Ctx` trait for attribute types. This would allow us to use a single `get` call with any type of attribute. This system currently has a few issues:

- ~~The `Attr` struct is currently used to parse attributes, but is pretty useless otherwise.~~
- ~~Ideally, we would like to be able to give different *sets* of attributes for different attribute contexts (ports vs components).~~


UnsignedByte:
@rachitnigam Currently attribute types are specified in the `utils` directory but that doesn't seem right. I could also move them to `ast` instead if that seems more reasonable but they are also used in IR...

rachitnigam:
I don't have strong opinions on where they should live! The `utils` separation is probably because they are included by both `ast` and `ir` but we don't want to include `ir` in ast.

UnsignedByte:
Yup, definitely agree with not including `ir` in AST. My main worry is that although the attribute traits and stuff could go into utils, the actual attributes themselves are somewhat tied to `Filament` and my impression of utils was that they should be "language-agnostic" compiler utilities... Maybe I will move attribs to AST.

rachitnigam:
No, the `utils` are utilities for the Filament compiler infrastructure. I don't think it is specifically architected to contained truly agnostic utilities.

UnsignedByte:
Macro-ized a bunch of stuff but I'm holding off on adding port attributes for another pass.

crates/ir/src/comp.rs
```diff
@@ -57,7 +58,7 @@ pub struct Component {
 
     // ============== Component signature ===============
     /// Attributes of the component
-    pub attrs: ast::Attributes,
+    pub attrs: utils::comp_attrs::Attrs,
```
rachitnigam, line 61: Any reason to expose the `comp_attrs` submodule?

UnsignedByte: Technically each set of `Attrs` comes with three types: `Bool, Num`, and `Attrs` for boolean and numeric flags as well as a type alias for the struct with these two types of flags.

We could publicize the underlying types instead, but that would then look something like `CompAttrs, CompBool, CompNum` and `PortAttrs, PortBool, PortNum` (for port attributes). Matter of preference honestly, I don't really mind either way.

crates/utils/src/attr/types.rs
```diff
@@ -0,0 +1,149 @@
+/*
+A macro to create a single attribute enum
```
rachitnigam, line 2: Can you add example of how it is used?

crates/utils/src/attr/types.rs
```diff
@@ -0,0 +1,149 @@
+/*
```
rachitnigam, line 1: Also, macros are visible at the crate level is usually good to put them in a centralized `macros.rs` file merely use them here.

rachitnigam:
Minor comments. Merge when resolved!

PR #475, opened 2024-10-29
Remove `PortDef` from AST

UnsignedByte:
Turns out `PortDef` is never used anymore (as everything is stored in bundles) so this cleans up that code and gets rid of `PortDef` entirely.


rachitnigam:
Ah wonderful! LGTM!

PR #477, opened 2024-10-31
Cleaning up monomorphize

UnsignedByte:
Cleaening up some of the monomorphize functions. @gabizon103 might find this interesting.

NOTE: `MonoSig` and `MonoDeferred` are unfortunately not going to be merged here (I tried, believe me) mainly because `MonoSig` is mutated all the time but much of `MonoDeferred` requires it to stay referenced, so merging the two causes a bunch of "mutably borrowed while alreaedy immutably borrowed" issues.


rachitnigam:
Is this ready for review? If so, LGTM!

PR #478, opened 2024-10-31
Update flopoco to latest

UnsignedByte:
The newest version of flopoco has DSP inference, so we are updating to it in order to possibly improve some of our results.


rachitnigam:
Let's disable the arm build stuff for now, at least for the `gen` stuff, while we work on the deadline.

PR #479, opened 2024-11-11
update frisc

gabizon103:
version on main was incomplete


apps/frisc/test/regfile.fil
```diff
@@ -1,18 +1,18 @@
...
-comp main<'G:1>(
+comp main<G:1>(
   clk:1,
-  reset: ['G,'G+1] 1,
-  en: ['G,'G+1] 1,
-  rd: ['G,'G+1] 5,
-  data: ['G,'G+1] 32,
-  rin1: ['G,'G+1] 5,
-  rin2: ['G,'G+1] 5
+  @[G,G+1] reset:1,
```
rachitnigam, line 5: This seems like old syntax that won't compile. Can we add the `frisc` tests to the test suite?

rachitnigam (changes requested):
Some of the files have the old syntax. Not sure if that is a merge problem or just out of date code. We should also add testing for this stuff to make sure things compile.

gabizon103:
I think something got messed up in one of the merges, thanks for pointing out. Planning on updating the tests before merging as well

PR #486, opened 2025-01-19
Port attributes

UnsignedByte:
Adds port attributes of the form

```
#[attrib]
port: ['G, 'G+1] 32,
...
```

Also can technically be used within bundle definitions like
```
bundle #[attrib] bun[3]: for<i> ['G + i, 'G+i+1] 32;
```

### DIFF
--- a/crates/ir/src/structure.rs
+++ b/crates/ir/src/structure.rs
@@ -3,6 +3,7 @@ use super::{
     InstIdx, InvIdx, ParamIdx, PortIdx, Subst, TimeIdx, TimeSub,
 };
 use fil_ast::Op;
+use fil_utils::PortAttrs;
 use itertools::Itertools;
 use std::fmt;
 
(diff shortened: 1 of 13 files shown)


UnsignedByte:
I should add a test case for this but right now there are no implemented attributes...

UnsignedByte:
@rachitnigam if there are no problems here I will merge this and start using it in the scheduling pass. My one question remaining is whether it is a good idea to store an `Attributes` in every port, considering we are trying to lower the size of the actual control elements. ALternatively we would have to create an `IndexStore` but I am not sure how reasonable that is.

rachitnigam:
There is a way to store information associated with a PortIdx that is not a port itself right? Something like a secondary context?

crates/ir/src/structure.rs
```diff
@@ -142,6 +143,7 @@ pub struct Port {
     pub width: ExprIdx,
     pub live: Liveness,
     pub info: InfoIdx,
+    pub attrs: PortAttrs,
```
rachitnigam, line 146: We should store this in a SparseIndexStore that maps PortIdx to PortAttrs

UnsignedByte: Great, exactly what I was looking for. Will do!

PR #488, opened 2025-01-24
Prevent double-definitions of propositions

### LINKED ISSUE #487: Propositions used in expressions are defined twice in `discharge`
Propositions that are used in expressions (namely in the ternary operation) are defined twice. This does not cause problems with `z3` (I assume it is overridden by the second definition, but as both are equal nothing bad happens), but `cvc5` has issues when this occurs.

https://github.com/cucapra/filament/blob/e0703dd72851f1d412ade4c6dc3b4b1ed791b159/crates/filament/src/ir_passes/discharge.rs#L619-L631

Every proposition is added again here:

https://github.com/cucapra/filament/blob/e0703dd72851f1d412ade4c6dc3b4b1ed791b159/crates/filament/src/ir_passes/discharge.rs#L673-L688

In CVC5 we get
```
[TRACE] -> (define-fun prop12 () Bool (> e37 e57))
[TRACE] <- (error "Parse Error: <shell>:1.13: Symbol 'prop12' previously declared as a variable")
```

UnsignedByte:
Fixes #487 

Also splits `check` tests into `cvc5` and `z3` to prevent things like this from occurring again. The existing `if-expr` test already checks this but was just never run on `cvc5`.

### DIFF
--- a/apps/blas/util.py
+++ b/apps/blas/util.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 import argparse
 from functools import reduce
 from itertools import chain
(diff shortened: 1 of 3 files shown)


rachitnigam (approved):
LGTM!

PR #489, opened 2025-01-24
add float attributes

UnsignedByte:


crates/utils/src/attr/attributes.rs
```diff
@@ -59,15 +64,40 @@ where
...
+where
+    Bool: FromStr + Hash + Eq + Copy,
+    Num: FromStr + Hash + Eq + Copy,
+    Float: FromStr + Hash + Eq + Copy,
 {
     fn default() -> Self {
         Self {
             num_attrs: AttrStore::default(),
             bool_attrs: AttrStore::default(),
+            float_attrs: AttrStore::default(),
```
rachitnigam, line 100: If you were really hankering for performance you could probably do some shenanigans to store floats and ints in the same array and only use 31-bits for the information and 1 bit for tracking the datatype

UnsignedByte: I don't think this matters in the long term especially if we want string attributes someday, mainly worried about code duplication right now lol, but macro-izing everything seems like a pain.

rachitnigam: Yeah, just an idle thought. String attributes would probably live on the heap behind a Box. But yeah, this looks fine.

PR #492, opened 2025-02-25
Associate attributes to passes

UnsignedByte:
Not sure if this is strictly necessary but with a lot of the work I am doing I need to have monomorphized components and it seems maybe a bit more explicit if we have attributes associated with these?


rachitnigam:
Not too terrible! I guess we're essentially using private attributes as fields on components.

PR #495, opened 2025-03-04
update toolchain version to 1.85

UnsignedByte:
`rust-analyzer` and a bunch of other stuff is starting to drop toolchain support for versions before 1.82, so this will update everything to 1.82 from 1.76 as that is probably what we will be developing on from now on anyway.


rachitnigam:
We could commit and move to 1.85 to get rust 2024 edition support.

UnsignedByte:
We now have to deal with these guys:
```
error: creating a mutable reference to mutable static is discouraged
  --> crates/utils/src/gsym.rs:22:9
   |
22 |         SINGLETON.assume_init_mut()
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable reference to mutable static
   |
   = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html>
   = note: mutable references to mutable statics are dangerous; it's undefined behavior if any other pointer to the static is used or if any other reference is created for the static while the mutable reference lives
```

This is described [here](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html), @rachitnigam do you have any idea on how you want to restructure this? Technically we could decide to pass everything around but I am not sure how much we want to do that, alternatively we could use raw pointers?

rachitnigam:
This is non-obvious. Curious what @EclecticGriffin did for Calyx which probably has the same problem?

UnsignedByte:
I've tried a bunch of stuff and nothing has really done the trick IMO - `RefCell` so far seems closest but it becomes impossible to get a `&'static str` reference from the string interner because `RefCell` generates a temporary `Ref` when borrowing. (for `GSym`)

```rust
fn singleton() -> &'static cell::RefCell<Pool> {
    static mut SINGLETON: mem::MaybeUninit<cell::RefCell<Pool>> =
        mem::MaybeUninit::uninit();
    static ONCE: sync::Once = sync::Once::new();

    // SAFETY:
    // - writing to the singleton is OK because we only do it one time
    // - the ONCE guarantees that SINGLETON is init'ed before assume_init_ref
    unsafe {
        ONCE.call_once(|| {
            SINGLETON.write(cell::RefCell::new(Pool::new()));
        });

        SINGLETON.assume_init_ref()
    }
}
// ...
impl From<GSym> for &'static str {
    fn from(sym: GSym) -> Self {
        singleton().borrow().resolve(sym.0).unwrap()
    }
}
```

UnsignedByte:
FWIW I think there is probably no "safe" way to do this anyway - there is no way that the compiler can know that a resolved static reference to a symbol in the pool will not be mutated (as long as more mutable references to the pool *could* still be made). Maybe we need some sort of builder construct...

UnsignedByte:
This PR disables `cargo build --all` in the dockerfile because Depot won't be able to build the main branch due to the rust version being changed. We could add this back in another PR after this is merged to main.

UnsignedByte:
Had to change the github workflow because actions like checkout run in `__w/filament/filament` regardless of what the default working directory is set to. This also technically means every test that has been run before now has running using the main branch's tests and not the tests in the dockerfile at all.

EclecticGriffin:
> This is non-obvious. Curious what @EclecticGriffin did for Calyx which probably has the same problem?

Oh, assuming this is for the symbol table, we ended up using `LazyLock` and `boxcar::Vec` rather than doing anything super fancy. You can take a look at some of the convo here: https://github.com/calyxir/calyx/pull/2377

There's a bunch of other stuff in the PR since we were also fixing CI version stuff, but you can ignore that

UnsignedByte:
This works great, thanks!

UnsignedByte:
@rachitnigam One problem is left in the file position table - we use `SimpleFiles` to represent the file map and it's used a lot in `codespan_reporting`. Not sure how we might want to replace this (or if we can) with boxcar?

UnsignedByte:
Done! No more mut statics.

PR #496, opened 2025-03-19
Move assumption/assertion generation to its own pass

UnsignedByte:
This PR divests core assumption/assertion generation (about event constraints, param constraints, and range constraints) from astconv. This cleans up astconv a bit and also adds the important ability for the IR to generate its own assertions necessary for discharging. This will be useful in scheduling and is also necessary if we want to be able to discharge after monomorphization properly.

### DIFF
--- a/crates/filament/src/ir_passes/fun_assumptions.rs
+++ b/crates/filament/src/ir_passes/fun_assumptions.rs
@@ -4,9 +4,9 @@ use fil_ir::{self as ir, AddCtx, Ctx, ExprIdx, PropIdx};
 
 /// Generates default assumptions to the Filament program for assumptions using custom functions
 #[derive(Default)]
-pub struct Assume;
+pub struct FunAssumptions;
 
-impl Assume {
+impl FunAssumptions {
     /// Adds the assumptions associated with a proposition of the form `#l = f(#r)` to the component.
     fn add_assumptions(
         ctx: &mut ir::Component,
@@ -56,7 +56,7 @@ impl Assume {
     }
 }
 
-impl Assume {
+impl FunAssumptions {
     /// Checks a proposition for whether it matches the form `#l = f(#r)` for some custom function `f`. Additionally recurses on `&` chains.
     /// Generates the assumptions associated with each [ast::Fn] and returns a list of [ir::Prop]s for each.
     /// TODO: Implement assumption generation for functions taking more than one argument.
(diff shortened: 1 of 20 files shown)


UnsignedByte:
@rachitnigam There is some finicky stuff here caused by the fact that `Visitor` does an `std::mem::take` on the component being visited, so in order to deal with recursive components we do some special checks.

crates/filament/src/ir_passes/bundle_elim.rs
```diff
@@ -4,25 +4,25 @@ use crate::{
...
+    PortIdx, PortOwner, Range, SparseInfoMap, Subst, Time,
 };
 use fil_utils as utils;
 use itertools::Itertools;
 use std::collections::HashMap;
 
-pub type PortInfo =
-    (/*lens=*/ Vec<usize>, /*gen_ports=*/ Vec<PortIdx>);
+/// A mapping from a bundle with a list of dimensions to the list of generated ports.
+type PortInfo = (/*lens=*/ Vec<usize>, /*gen_ports=*/ Vec<PortIdx>);
```
rachitnigam, line 15: If we're extracting these into separate types, might as well give the fields real names?

crates/filament/src/ir_passes/fun_assumptions.rs
```diff
@@ -4,9 +4,9 @@ use fil_ir::{self as ir, AddCtx, Ctx, ExprIdx, PropIdx};
 
 /// Generates default assumptions to the Filament program for assumptions using custom functions
 #[derive(Default)]
-pub struct Assume;
+pub struct FunAssumptions;
```
rachitnigam, line 7: Nit: I don't love the name. Can we use something like `InferAssumes` which implies some relationship between existing assumptions and created ones?

crates/ir/src/utils/transfer.rs
```diff
@@ -0,0 +1,206 @@
+use super::{Bind, Foreign};
+use crate::{
+    AddCtx, CmpOp, Ctx, Event, Expr, ExprIdx, MutCtx, Param, Prop, PropIdx,
+    Time, TimeIdx, TimeSub,
+};
+
+impl<C> Foreign<Expr, C>
```
rachitnigam, line 7: What is this exactly used for and can we give this a better name/trait wrapper?

rachitnigam (approved):
Seems cool! I'm curious if this basically brings us back to the old world where the well-formed pass would add all of the assumptions and they could be discharged later?

UnsignedByte:
> Seems cool! I'm curious if this basically brings us back to the old world where the well-formed pass would add all of the assumptions and they could be discharged later?

Yes this is the goal hopefully, the nice thing about this is we can now on-the-fly generate all the assertions/assumptions that are important and then something like scheduling can just work directly off these.

crates/ir/src/utils/transfer.rs
rachitnigam: Got it, that seems reasonable to me!

rachitnigam:
Awesome! Merge whenever!

PR #499, opened 2025-04-29
Fix getting started instructions

### LINKED ISSUE #451: update docs for how to test the Filament installation
### problem
the "getting started" page says to run `cargo run -- tests/compile/par.fil` to test the build.

doing so I get a few errors:
```sh
$ cargo run -- ./tests/compile/par.fil
error: `cargo run` could not determine which binary to run. Use the `--bin` option to specify a binary, or the `default-run` manifest key.
available binaries: fil-gen, filament
$ cargo run --bin=filament -- ./tests/compile/par.fil
warning: ...
warning: ...

Error: invalid file: Failed to read ./tests/compile/par.fil: No such file or directory (os error 2): 
Compilation failed with 1 errors.
Run with --show-models to generate assignments for failing constraints.
```

I can't find a file named `./tests/compile/par.fil` either.

### the request
it seems that the docs should:
1. include the `--bin=filament` flag, and
2. list a working test file (or set of test files)
    - it would be great too if there were a command to run all tests at once, to make the process as simple as possible

Timmmm:
The instruction suggest using a file that no longer exists. I tested this file and it works.

Fixes #451

### DIFF
--- a/doc/docs/start.md
+++ b/doc/docs/start.md
@@ -17,7 +17,7 @@ Next, we can install the dependencies for the Filament compiler:
 
 To check that the compiler works, run the following command:
 ```
-cargo run -- tests/compile/par.fil
+cargo run -- tests/run/add.fil
 ```
 Which should generate the Verilog implementing the original program.
 


rachitnigam:
LGTM, thanks!

PR #507, opened 2025-08-08
Warnings for unused bindings

rachitnigam:
Produce warnings when particular kinds of bindings are not used in the program.

### DIFF
--- a/tests/warnings/unused/simple-unused.expect
+++ b/tests/warnings/unused/simple-unused.expect
@@ -1,23 +1,35 @@
 ---CODE---
 1
 ---STDERR---
-error: unused instance: instance is created but never invoked
-  ┌─ tests/warnings/unused/simple-unused.fil:8:3
+error: unused parameter: parameter is defined but never used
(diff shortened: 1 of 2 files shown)


crates/filament/src/ast_passes/unused_elements.rs
```diff
@@ -0,0 +1,421 @@
...
+    instances: HashMap<String, ast::Loc<fil_utils::Id>>,
+    invocations: HashMap<String, ast::Loc<fil_utils::Id>>,
+    let_params: HashMap<String, ast::Loc<fil_utils::Id>>,
+    sig_params: HashMap<String, ast::Loc<fil_utils::Id>>,
+
+    // Usage tracking
+    invoked_instances: HashSet<String>,
+    accessed_invocations: HashSet<String>,
+    referenced_params: HashSet<String>,
+}
```
rachitnigam, line 20: This approach tracks one single scope for the entire component but constructs like "for" and "if" create new scopes and bindings may shadow each other. This means that the single flat global scope would not work.

